PR #55, opened 2017-08-14
Remove PhysConst class

hansgans:
-Remove PhysConst class and use instead map of ParticleProperties
-fix XML read and write of model


hansgans:
#55

PR #65, opened 2017-10-23
Various changes

hansgans:
- Build system: fixing install target
- CI: improve travisCI integration and coverity_scan
- Physics: Simultaneous likelihood

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ doc/doxygen/
 build.cfg
 /Default/
 build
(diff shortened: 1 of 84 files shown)


MathiasMichel:
in MinuitResult.cpp, replace line 212 with:
    try{
      *tableCorr << globalCC.at(n);
    }catch(...){
      *tableCorr << "?";
    }
to catch exception when global correlation is not valid

PR #76, opened 2018-03-13
Canonical basis

hansgans:


zhang-jingqing:
The Faddeeva is just used to calculate the Voigtian function, which is the result of non-relativisitic Breit-Wigner convolved with a Gaussian (the detector resolution in the mass spectrum). If one doesn't use the Voigtian in the Decay, it does nothing. The Faddeeva things only related to the new Voigtian class in DecayDynamics. And it is not part of the HelicityFromCanonicalSum.
The HelicityFromCanonicalSum will just transfer one helicity decay amplitude to the a sum of the canonical decay amplitudes, and at last, these decay amplitudes will be added to the up-level, SquentialPartialDecayAmplitude separately. 
And I also fixed some bugs for these two implementations. Now they can run successfully.
I will add some descriptions after I have a look for how to add it in doxygen...

PR #79, opened 2018-05-11
Expert system

spflueger:
This pull request adds a first version of the new python expert system to ComPWA. Also some code refactoring in form of design optimizations and code cleanup is included.


Core/FitParameter.hpp
```diff
@@ -20,14 +20,51 @@
...
 
+struct BoolTranslator {
+  typedef std::string internal_type;
+  typedef bool external_type;
+
+  // Converts a string to bool
+  boost::optional<external_type> get_value(const internal_type &str) {
+    if (!str.empty()) {
+    	using boost::algorithm::iequals;
+      if (iequals(str, "true") || iequals(str, "yes") || str == "1")
```
hansgans: We could add "True" here

Core/SubSystem.cpp
```diff
@@ -71,86 +74,100 @@ boost::property_tree::ptree SubSystem::save() const {
...
-      stream << "_";
+  for (auto const& j : _finalStates) {
+    for (auto const& i : j) stream << std::to_string(i);
+    stream << "_";
   }
-  stream << "_vs_";
-  for (auto i : _recoilState)
-    stream << std::to_string(i);
+  stream << "vs_";
+  for (auto i : _recoilState) stream << std::to_string(i);
```
hansgans: Some parts of the diff is only due to a line break after control statements. I didn't explicitly find it in the LLVM standard but all the examples have a line break.

Core/SubSystem.cpp
```diff
@@ -71,86 +74,100 @@ boost::property_tree::ptree SubSystem::save() const {
...
 
   return stream.str();
 }
 
-bool SubSystem::operator==(const SubSystem &b) const {
+bool SubSystem::operator==(const SubSystem& b) const {
```
hansgans: I think LLVM suggests `SubSystem &b` over  `SubSystem& b`

Core/FitParameter.hpp
spflueger: That is already included. It uses iequals, which is a case insensitive compare. So TRUE True TRue true YES Yes yes etc... and 1 all works

PR #82, opened 2018-06-05
correct helicity angle implementation and corresponding unit test

spflueger:


MathiasMichel:
I get a compile error from DalitzPlot because of roots TString using c++17's "std::string_view". might be something locally on my machine, but in case somebody else got similar effects i note it here. I will try to figure out what happens.

PR #90, opened 2018-08-07
update the usage of toySample. It is expected to do the

zhang-jingqing:
update the usage of toySample. It is expected to do the
normalization for amplitudes(e.g., BreitWigner and WignerD).
But actually, the phspSample is used instead of toySample.
The difference between phspSample and toySample is
that ususally phspSample will include the efficiency(i.e.,
only events passed selection) while toySample not.

### DIFF
--- a/Physics/CoherentIntensity.cpp
+++ b/Physics/CoherentIntensity.cpp
@@ -129,7 +129,7 @@ CoherentIntensity::tree(std::shared_ptr<Kinematics> kin,
 
   for (auto i : Amplitudes) {
     std::shared_ptr<ComPWA::FunctionTree> resTree =
-        i->tree(kin, sample, phspSample, "");
+        i->tree(kin, sample, toySample, "");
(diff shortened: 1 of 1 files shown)


spflueger:
Hmm, i dont really understand this matter. Afaik the phspSample should be used for the normalization. I'm not sure why toy sample exists here in the first place? What is that really needed for?

hansgans:
Thanks for spotting the bug. 
@spflueger As @zhang-jingqing  said the difference between phspSample and toySample is that the latter one does not include the efficiency. When you calculate the normalization of a BW you don't want to include the efficiency. On the contrary, the normalization integral of the LH should take the efficiency into account.

spflueger:
Hmm, yeah that make sense, but I'm having troubles with the names of the data samples. I would expect a phsp sample to be "phase space", and not to include efficiency correction. Can't we rename toySample to phspSampleWithEffciency and change things according to that?

PR #158, opened 2019-03-26
Added Minuit2 Optimization Test again

### LINKED ISSUE #157: Optimizer tests are commented out
Seems that test routines needs to be adapted to the recent changes

spflueger:
- Closes #157 (Geneva tests remain switched off for now, since the CI
  does not include the Geneva package)
- Correctly implemented an optimization test for the Minuit2 interface
  using root finding of several functions


spflueger:
I want to implement the geneva stuff as well. However that is not working out of the box as I want to. Maybe I will get it finished by tomorrow.

PR #162, opened 2019-04-01
Production formfactor

zhang-jingqing:
Update: the production formfactor is implemented in a decorator patern. The FormFactorDecorator
will decorate a BreitWigner with a BlattWeisskopf formfactor(for the production the decay process,
R -> a b). Note, the decay formfactor which appears in Gamma(s) of a relativisitic Breit-Wigner, 
is implemented in the original BreitWigner, and it is not related to this FormFactorDecorator class 
currently.
This FormFactorDecorator will take two additional FitParameters than the original BW, the masses 
of two daughters. It is useful in q sequential decay where we also want to fit the intermediate resonances' mass. e.g., J/psi -> gamma f_0, f_0 -> pi0 pi0, then we can take mass of f_0 as fit parameter and for the FormFactorDecorator, it will work well for the two sub decays.

### DIFF
--- a/Physics/Dynamics/ProductionFormFactor.cpp
+++ b/Physics/Dynamics/ProductionFormFactor.cpp
@@ -16,28 +16,17 @@ namespace Physics {
 namespace Dynamics {
 
 ProductionFormFactor::ProductionFormFactor(std::string name,
-                                           std::string daug1Name,
-                                           std::string daug2Name,
                                            std::shared_ptr<ComPWA::FitParameter> mass1,
                                            std::shared_ptr<ComPWA::FitParameter> mass2,
                                            std::shared_ptr<ComPWA::FitParameter> radius,
                                            ComPWA::Spin orbitL,
                                            FormFactorType ffType)
-    : AbstractDynamicalFunction(name) {
+    : AbstractDynamicalFunction(name),
+    Daughter1Mass(mass1), Daughter2Mass(mass2), MesonRadius(radius),
+    L(orbitL), FFType(ffType) {
 
-  LOG(TRACE) << "ProductionFormFactor::Factory() | Construction of " << name
-             << ".";
-
-  SetDaughter1MassParameter(mass1);
-  SetDaughter2MassParameter(mass2);
-  SetMesonRadiusParameter(radius);
-  SetOrbitalAngularMomentum(orbitL);
-  SetFormFactorType(ffType);
-
-  LOG(TRACE)
-      << "ProductionFormFactor::Factory() | Construction of the decay "
-      << name << " -> " << daug1Name << " + "
-      << daug2Name;
+  LOG(TRACE) << "ProductionFormFactor::Factory() | Construction production "
+             << "formfactor of " << name << ".";
 }
 
 ProductionFormFactor::~ProductionFormFactor() {}
--- a/Physics/Dynamics/ProductionFormFactor.cpp
+++ b/Physics/Dynamics/ProductionFormFactor.cpp
@@ -211,6 +211,7 @@ void ProductionFormFactor::updateParametersFrom(const ParameterList &list) {
   if (rad)
     MesonRadius->updateParameter(rad);
 
+  // Try to update daugher1's mass
   std::shared_ptr<FitParameter> daug1Mass;
   try{
     daug1Mass = FindParameter(Daughter1Mass->name(), list);
@@ -219,6 +220,7 @@ void ProductionFormFactor::updateParametersFrom(const ParameterList &list) {
   if (daug1Mass)
     daug1Mass->updateParameter(daug1Mass);
 
+  // Try to update daugher2's mass
   std::shared_ptr<FitParameter> daug2Mass;
   try{
     daug2Mass = FindParameter(Daughter2Mass->name(), list);
--- a/Physics/Dynamics/ProductionFormFactor.cpp
+++ b/Physics/Dynamics/ProductionFormFactor.cpp
@@ -1,0 +1,248 @@
+// Copyright (c) 2013, 2017 The ComPWA Team.
+// This file is part of the ComPWA framework, check
+// https://github.com/ComPWA/ComPWA/license.txt for details.
+
+#include <cmath>
+#include <iterator>
+#include <numeric>
+#include <vector>
+
+#include "Coupling.hpp"
+#include "Physics/HelicityFormalism/HelicityKinematics.hpp"
+#include "ProductionFormFactor.hpp"
+
+#include <boost/property_tree/ptree.hpp>
+
+namespace ComPWA {
+namespace Physics {
+namespace Dynamics {
+
+ProductionFormFactor::ProductionFormFactor(
+    std::string name, std::pair<std::string, std::string> daughters,
+    std::shared_ptr<ComPWA::PartList> partL)
+    : AbstractDynamicalFunction(name) {
+
+  LOG(TRACE) << "ProductionFormFactor::Factory() | Construction of " << name
+             << ".";
+
+  auto partProp = partL->find(name)->second;
+
+  auto decayTr = partProp.GetDecayInfo();
+
+  auto spin = partProp.GetSpinQuantumNumber("Spin");
+  J = spin;
+  // in default, using spin J as Orbital Angular Momentum
+  // update by calling SetOrbitalAngularMomentum() before any further process
+  // after the ProductionFormFactor is created by calling of constructor
+  L = spin;
+
+  FFType = FormFactorType(decayTr.get<int>("FormFactor.<xmlattr>.Type"));
+
+  // Read parameters from tree. Currently parameters of type 'MesonRadius' is required.
+  // But width is also in this tree, because it is needed by RelBW
+  // IF non-Resonant, they will be not decay Info,
+  for (const auto &v : decayTr.get_child("")) {
+    if (v.first != "Parameter")
+      continue;
+    std::string type = v.second.get<std::string>("<xmlattr>.Type");
+    if (type == "MesonRadius") {
+      SetMesonRadiusParameter(std::make_shared<FitParameter>(v.second));
+    } else {
+      if (type != "Width") {
+        throw std::runtime_error(
+            "ProductionFormFactor::Factory() | Parameter of type " + type +
+            " is unknown.");
+      }
+    }
+  }
+
+  DaughterMasses =
+      std::make_pair(partL->find(daughters.first)->second.GetMass(),
+                     partL->find(daughters.second)->second.GetMass());
+
+  DaughterNames = daughters;
+
+  LOG(TRACE)
+      << "ProductionFormFactor::Factory() | Construction of the decay "
+      << partProp.name() << " -> " << daughters.first << " + "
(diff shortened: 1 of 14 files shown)


Physics/Dynamics/ProductionFormFactor.hpp
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2013, 2019 The ComPWA Team.
+// This file is part of the ComPWA framework, check
+// https://github.com/ComPWA/ComPWA/license.txt for details.
+
+#ifndef PHYSICS_DYNAMICS_PRODUCTION_FORMFACTOR_HPP_
+#define PHYSICS_DYNAMICS_PRODUCTION_FORMFACTOR_HPP_
+
+#include <boost/property_tree/ptree.hpp>
```
spflueger: This class should not depend on the boost property tree. All of the property tree code should go into the IntensityBuilderXML file

Physics/Dynamics/ProductionFormFactor.hpp
```diff
@@ -0,0 +1,130 @@
...
+  void addUniqueParametersTo(ParameterList &list);
+  void addFitParametersTo(std::vector<double> &FitParameters) final;
+
+  std::shared_ptr<FunctionTree>
+  createFunctionTree(const ParameterList &DataSample, unsigned int pos,
+                     const std::string &suffix) const;
+
+protected:
+  /// Resonance spin
+  ComPWA::Spin J;
```
spflueger: This variable is not needed. Only use L.

Physics/Dynamics/ProductionFormFactor.hpp
zhang-jingqing: Yes, you are right.

Physics/Dynamics/ProductionFormFactor.hpp
zhang-jingqing: Yes, I will update it

Physics/Dynamics/ProductionFormFactor.hpp
```diff
@@ -0,0 +1,130 @@
...
+  /// Orbital Angular Momentum between two daughters in Resonance decay
+  ComPWA::Spin L;
+
+  /// Masses of daughter particles
+  std::pair<double, double> DaughterMasses;
+
+  /// Names of daughter particles
+  std::pair<std::string, std::string> DaughterNames;
```
spflueger: I think this is also not needed.

Physics/Dynamics/ProductionFormFactor.hpp
```diff
@@ -0,0 +1,130 @@
...
+/// Ref. Dalitz Plot Analysis, PDG 2012, where 
+/// \f[
+/// \frac{B^{\prime}(q_{0})}{B^{\prime}_{L}(q)} = q^{L}B^{\prime}_{L}(q, q_{0})
+/// \f]
+class ProductionFormFactor
+    : public ComPWA::Physics::Dynamics::AbstractDynamicalFunction {
+
+public:
+  //============ CONSTRUCTION ==================
+  ProductionFormFactor(std::string name,
```
spflueger: Try to make the constructor get all of the variables it need to be fully constructed. In this case a L, a pair of masses, meson radius, form factor type. All of the xml extraction should go into the XMLBuilder

Physics/Dynamics/ProductionFormFactor.hpp
```diff
@@ -0,0 +1,130 @@
...
+                          std::pair<std::string, std::string> daughters,
+                          std::shared_ptr<ComPWA::PartList> partL);
+  virtual ~ProductionFormFactor();
+
+  boost::property_tree::ptree save() const;
```
spflueger: This function was removed. Don't implement it

Physics/Dynamics/ProductionFormFactor.hpp
zhang-jingqing: We need the daughers'mass to calcualte the breakup momentum, q, which is used to calculate the Blatt-Weisskopf F_L(q)

Physics/Dynamics/ProductionFormFactor.hpp
zhang-jingqing: Yes, I just removed it.

Physics/Dynamics/ProductionFormFactor.hpp
spflueger: Btw I think its good if we introduce a new XML tag for the form factors `<FormFactor>`, which contains all of the necessary information

zhang-jingqing: OK. That is good idea, I will try it.

Physics/Dynamics/ProductionFormFactor.hpp
spflueger: But why the names of the daugther masses?

zhang-jingqing: Use the names to extract the masses in PartilePropertyList...

spflueger: Yeah, but that code should go into the XMLBuilder again

Physics/Dynamics/ProductionFormFactor.hpp
```diff
@@ -0,0 +1,155 @@
...
+  /// \param mb Mass of daughter particle
+  /// \param L Orbital angular momentum between two daughters a and b
+  /// \param mesonRadius Meson Radius
+  /// \param ffType Form factor type
+  static std::complex<double>
+  dynamicalFunction(double mSq, double ma, double mb, unsigned int L, 
+                    double mesonRadius, FormFactorType ffType);
+
+  //============ SET/GET =================
+
```
spflueger: Are these getters and setters really necessary? I think you can remove all of them.

zhang-jingqing: emm, you are right

Physics/Dynamics/ProductionFormFactor.hpp
```diff
@@ -0,0 +1,155 @@
...
+  void updateParametersFrom(const ParameterList &list);
+  void addUniqueParametersTo(ParameterList &list);
+  void addFitParametersTo(std::vector<double> &FitParameters) final;
+
+  std::shared_ptr<FunctionTree>
+  createFunctionTree(const ParameterList &DataSample, unsigned int pos,
+                     const std::string &suffix) const;
+
+protected:
```
spflueger: make them private instead of protected

zhang-jingqing: ok

Physics/Dynamics/ProductionFormFactor.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include "Coupling.hpp"
+#include "Physics/HelicityFormalism/HelicityKinematics.hpp"
+#include "ProductionFormFactor.hpp"
+
+namespace ComPWA {
+namespace Physics {
+namespace Dynamics {
+
+ProductionFormFactor::ProductionFormFactor(std::string name,
+                                           std::string daug1Name,
```
spflueger: remove daug1/2Name. They are not used

Physics/Dynamics/ProductionFormFactor.cpp
```diff
@@ -0,0 +1,237 @@
...
+                                           std::shared_ptr<ComPWA::FitParameter> mass2,
+                                           std::shared_ptr<ComPWA::FitParameter> radius,
+                                           ComPWA::Spin orbitL,
+                                           FormFactorType ffType)
+    : AbstractDynamicalFunction(name) {
+
+  LOG(TRACE) << "ProductionFormFactor::Factory() | Construction of " << name
+             << ".";
+
+  SetDaughter1MassParameter(mass1);
```
spflueger: initialize the members via `ProductionFormFactor::ProductionFormFactor(...) : ..... {}`

Physics/HelicityFormalism/HelicityDecay.hpp
```diff
@@ -49,6 +51,9 @@ class HelicityDecay : public NamedAmplitude {
   std::shared_ptr<ComPWA::Physics::Dynamics::AbstractDynamicalFunction>
       DynamicFunction;
 
+  std::shared_ptr<ComPWA::Physics::Dynamics::AbstractDynamicalFunction>
```
spflueger: I would remove this (actually all of you changes in the HelicityDecay class), and instead go for the decorator pattern.

Physics/Dynamics/ProductionFormFactor.hpp
```diff
@@ -0,0 +1,98 @@
...
+/// \f[
+/// \frac{B^{\prime}(q_{0})}{B^{\prime}_{L}(q)} = q^{L}B^{\prime}_{L}(q, q_{0})
+/// \f]
+/// \f$B_{L}(q)(i.e., F_L(q))$\f with \f$L$\f up to 4 can be found in BNL-QGS-06-101
+class ProductionFormFactor
+    : public ComPWA::Physics::Dynamics::AbstractDynamicalFunction {
+
+public:
+  //============ CONSTRUCTION ==================
+  ProductionFormFactor(std::string name,
```
spflueger: With the decorator pattern you would pass a shared_ptr of the "undecorated" class here. That would be just another AbstractDynamicalFunction. So `std::shared_ptr<AbstractDynamicalFunction>`. Then you also define a member in this class which you assign this to. Finally you only need to modify the `evaluate()` function as shown below

Physics/Dynamics/ProductionFormFactor.cpp
```diff
@@ -0,0 +1,226 @@
...
+    Daughter1Mass(mass1), Daughter2Mass(mass2), MesonRadius(radius),
+    L(orbitL), FFType(ffType) {
+
+  LOG(TRACE) << "ProductionFormFactor::Factory() | Construction production "
+             << "formfactor of " << name << ".";
+}
+
+ProductionFormFactor::~ProductionFormFactor() {}
+
+std::complex<double> ProductionFormFactor::evaluate(const DataPoint &point,
```
spflueger: Here you would call the "undecorated" member's `evaluate()` function and multiply this class `dynamicalFunction()` function on top. Maybe even rename `dynamicalFunction()` here to something more meaningful.

spflueger:
@zhang-jingqing The codecov should not fail now. Unfortunately you have to trigger the checks again by pushing to you pull request.

Physics/Dynamics/ProductionFormFactorDecorator.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+#include "Coupling.hpp"
+#include "Physics/HelicityFormalism/HelicityKinematics.hpp"
+#include "ProductionFormFactorDecorator.hpp"
+
+namespace ComPWA {
+namespace Physics {
+namespace Dynamics {
+
+ProductionFormFactorDecorator::ProductionFormFactorDecorator(std::string name,
```
spflueger: I would rename the class to just `FormFactorDecorator`, since you can use this class not only in the production.

PR #174, opened 2019-04-12
Tools for update ptree parameters

zhang-jingqing:
Provide some tools to update a property tree's parameters (value, range, fix or not).

With the serialization method of MinuitResult, one can save and load the fit result.
And use these tools to update the property tree's parameters  according the loaded
fit result. Then one can use this updated property tree to build a new Intensity object,
and to calculate the fit fractions for the loaded fit result.  A test file Tools/test/SaveAndLoadFitResultTest.cpp is used to show how it works.

### DIFF
--- a/Tools/CMakeLists.txt
+++ b/Tools/CMakeLists.txt
@@ -6,11 +6,11 @@
 if( ${ROOT_FOUND} AND ${GSL_FOUND})
 
 set(lib_srcs
-  EvtGenGenerator.cpp Generate.cpp RootGenerator.cpp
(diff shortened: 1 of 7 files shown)


spflueger:
Hi @zhang-jingqing,
can you get rid of your second commit, the merge commit. I want to try to have a bit cleaner commit history in the future. So just merge those two commits via `rebase` or do a `rebase` to the `upstream/master` and remove that second commit.

hansgans:
That looks good. Could you probably make those methods available in the pybind interface?

zhang-jingqing:
@hansgans OK. I will do it next week

PR #175, opened 2019-04-17
Fixes in pycompwa and improvements in expert system

spflueger:


pycompwa/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ if(${Python3_FOUND})
 
   # Copy to build directory
   configure_file(${CMAKE_CURRENT_SOURCE_DIR}/setup.py.in ${SETUP_PY})
-  add_custom_command(OUTPUT ${OUTPUT}
+  add_custom_target(pycompwa ALL
```
hansgans: This is not the same as the previous implementation. Please revert.

pycompwa/pycompwa/plotting/__init__.py
```diff
@@ -1,6 +1,5 @@
 import matplotlib.pyplot as plt
-# from matplotlib import rc
-from matplotlib.colors import LinearSegmentedColormap
+
```
hansgans: I would like to avoid modules containing a single file

spflueger: I forgot to add the `rootplotdatareader.py` file, I will update the pull request shortly

pycompwa/CMakeLists.txt
spflueger: Hmm, I had some problems of python files being copied when I made modifications. I will check that again

spflueger: Yeah, I just tried it again. When I make some modifications to some python code (after the build), and I build again, the modifications are not copied to the build directory. Why do you need this `add_custom_command`?

spflueger:
Ok, now everything should be included. If you are happy with the pull request you can merge it.

PR #177, opened 2019-04-25
fix a bug in FormFactorDecorater

zhang-jingqing:
This bug makes the production form factor not applied.
Add some test to check if production form factor is really used as expected.


spflueger:
Hi @zhang-jingqing 
can you update this pull request so we can merge it

zhang-jingqing:
This is a hot fix to include the production formfactor in the amplitudes. The current implementation of production formfactor is in a decorator pattern. We can change it if we decide to implement it in another pattern/style in future.

PR #187, opened 2019-05-15
new fit result struct proposal

spflueger:
This is my proposal for the rework of the fit result classes. Its not finished, but it should give an idea how it works. If we decide to use this I will finish this up

### DIFF
--- a/Core/FitResult.cpp
+++ b/Core/FitResult.cpp
@@ -7,7 +7,7 @@
 #include "Core/Logging.hpp"
 
 namespace ComPWA {
-
+/*
 void FitResult::writeText(std::string filename) {
   std::ofstream myfile;
   myfile.open(filename, std::ios::app);
@@ -230,6 +230,6 @@ void FitResult::printFitFractions(TableFormater *fracTable) {
   SumFractionsError = sqrt(sumErrorSq);
 
   return;
-}
+}*/
 
 } // namespace
--- a/Core/FitResult.hpp
+++ b/Core/FitResult.hpp
@@ -2,137 +2,394 @@
 // This file is part of the ComPWA framework, check
 // https://github.com/ComPWA/ComPWA/license.txt for details.
 
-///
-/// \file
-/// Base class FitResult.
-///
+#ifndef COMPWA_FITRESULT_HPP_
+#define COMPWA_FITRESULT_HPP_
 
-#ifndef _FITRESULT_HPP_
-#define _FITRESULT_HPP_
-
-#include <vector>
-#include <string>
-#include <iostream>
-#include <fstream>
-#include <memory>
-
-#include <boost/serialization/serialization.hpp>
-#include <boost/serialization/nvp.hpp>
-#include <boost/serialization/vector.hpp>
-#include <boost/serialization/assume_abstract.hpp>
-#include <boost/serialization/export.hpp>
+#include <unordered_map>
 
 #include "Core/ParameterList.hpp"
-#include "Core/TableFormater.hpp"
-#include "Core/Properties.hpp"
-#include "Intensity.hpp"
+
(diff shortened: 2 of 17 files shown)


spflueger:
Here is a list of pro and cons of this approach:
Pro:
- `FitResult` is a pure data structure, which means that adding new functions working with `FitResult` structs is easy. Theoretically, the downside of pure data structures are that adding new members to `FitResult` would require the modification of all functions working with `FitResult`. However, the `std::map` structure allows for flexibility of adding new members, and only changes in functions where these new members are required. So this downside is pretty much eliminated (especially if one uses std::any/std::variant but that requires c++17)
- possibility to dynamically add information/properties to the `FitResult`
- only information, which is actually present is inside the `FitResult` (no empty/uninitialized fields)
- All optimizers return the same `FitResult` structure, which does not require any casting
- All optimizer instances only have to fill the `FitResult` and do not have to create a new inherited `FitResult` (avoids overuse of inheritance)

Cons:
- Compiler cannot detect wrong access to the `std::map` containers, which can result in runtime errors or weird behavior, unless some safety feature like a getter and setter are used.

zhang-jingqing:
It sounds good. I hope I can access parameters' name/values/errors/error matrix 
and fit status (edm, -logL, minimum valid or not, converged or not, error matrix valid or not etc.).
I think we also need to care how to load a such fit result to the specific in used, e.g., load fit result to recover a minuit2 fit result. (I am not sure if this feature will be used or not. If it will be never used, 
we can skip this).

spflueger:
Do you think it would be better to use inheritance here? So we have a base class `FitResult` and add more properties as needed by inheriting from that.
And do you think it is still comprehensible or transparent to a developer and user what is stored inside the FitResult?

hansgans:
I do not agree with the proposed change. The huge disadvantage is that the content of `std::map` in `FitResult.hpp` is not known at compile time. Which has several disadvantages:
* A proper writen CPP application would need to encapsulate each call to those `std::map`'s in a `try...catch` statement (that might be fine in python)
* A dynamical approach is not nessessary since the optimizer interfaces is not dynamic
* Given you write CPP code. Where do you find the information whats under which name is stored? It is not automatically documented via doxygen and you can not document it in the `FitResults.hpp`, the place where you would expect such an information. Since the content is not known to the compiler you can not use tab completition in your IDE.

So I propose to stick with an inheritance pattern:
```
class FitResult
{
    inital, final parameters
}
class MinuitResult : public FitResult 
{
    minos, hesse information
}
```

spflueger:
> A proper writen CPP application would need to encapsulate each call to those std::map's in a try...catch statement (that might be fine in python)

I don't see that as a "huge" disadvantage.

> A dynamical approach is not nessessary since the optimizer interfaces is not dynamic

This is not a disadvantage. If its anything it is an advantage.

I definitely prefer the `struct` solution with `std::map`, for the reasons I listed above (and there might be even more). Let me make a small example of a disadvantage of the inheritance solution:

Let's say there is some property `X`, which is not inside the base class `FitResult`. If there is some function which performs some task using this property `X`, you need to give this function an instance of inherited FitResults which have this property `X` defined. However if this property `X` is defined in many different inherited FitResult types, then you need to copy this function for each of these types.
You can of course decide to make this property `X` something "common" and put it inside the base class. But then some Optimizers might not use this property and it will be unset. This is not good and a nice example of when **not** to use inheritance (for example: https://softwareengineering.stackexchange.com/questions/12439/code-smell-inheritance-abuse). So you are stuck with writing multiple functions.

Also since we expose the `FitResult` to the python interface, it is much easier if this is just a simple data structure like proposed here (I think I don't need to elaborate this part)

hansgans:
The first answer in your post suggests inheritance for a comparable problem.

spflueger:
Was solved in #197

PR #197, opened 2019-07-10
Interface improvements

spflueger:


spflueger:
I still have to make two modifications (because the angular distributions are not correct now, so also two python tests will fail). But since these changes are small, I already made the pull request so you can have a look at the new changes.

Core/FunctionTree/FunctionTree.cpp
```diff
@@ -2,19 +2,20 @@
 // This file is part of the ComPWA framework, check
 // https://github.com/ComPWA/ComPWA/license.txt for details.
 
-#include "Core/FunctionTree.hpp"
+#include "FunctionTree.hpp"
```
hansgans: Shouldn't that be `Core/FunctionTree/FunctionTree.hpp`?

Core/FunctionTree/FunctionTreeEstimatorWrapper.hpp
```diff
@@ -0,0 +1,38 @@
...
+  std::shared_ptr<FunctionTree> getFunctionTree() { return Tree; }
+  ParameterList &getParameterList() { return Parameters; };
+  ParameterList &getUserParameterList() { return UserParameters; };
+
+  std::string print(int level) const;
+
+private:
+  std::shared_ptr<FunctionTree> Tree;
+  ParameterList Parameters;
+  ParameterList UserParameters;
```
hansgans: What is the difference between `Parameters` and `UserParameters`?

Core/FunctionTree/FunctionTreeIntensityWrapper.cpp
```diff
@@ -0,0 +1,89 @@
...
+FunctionTreeIntensityWrapper::FunctionTreeIntensityWrapper(
+    std::shared_ptr<OldIntensity> oldintens,
+    std::shared_ptr<ComPWA::Kinematics> kin, std::string name)
+    : FunctionTreeIntensityWrapper(
+          oldintens, kin->getKinematicVariableNames().size(), name) {}
+
+std::vector<double> FunctionTreeIntensityWrapper::evaluate(
+    const std::vector<std::vector<double>> &data) {
+  updateDataContainers(data);
+  Tree->UpdateAll(Tree->head());
```
hansgans: Does that mean that the `FunctionTree` is recalculated completely each time the data vectors have changed?

Core/FunctionTree/FunctionTreeIntensityWrapper.cpp
```diff
@@ -0,0 +1,89 @@
...
+void FunctionTreeIntensityWrapper::updateDataContainers(
+    const std::vector<std::vector<double>> &data) {
+  // just loop over the vectors and fill in the data
+  if (Data.mDoubleValues().size() > data.size()) {
+    std::stringstream ss;
+    ss << "FunctionTreeIntensityWrapper::updateDataContainers(): given data "
+          "container does not have enough variables! (required: "
+       << Data.mDoubleValues().size() << ", given: " << data.size() << ")";
+    throw std::out_of_range(ss.str());
+  }
```
hansgans: What if `(Data.mDoubleValues().size() < data.size()`?

Core/FunctionTree/ParameterList.hpp
```diff
@@ -78,91 +82,86 @@ class ParameterList {
 
   // Value
   // Single sized values
-  virtual std::shared_ptr<ComPWA::Value<int>> intValue(size_t i) {
+  virtual std::shared_ptr<Value<int>> intValue(size_t i) {
```
hansgans: In the header files we should stick to the ful namespace.

Core/FunctionTree/TreeNode.hpp
```diff
@@ -10,16 +10,16 @@
...
 #include <memory>
+#include <string>
 
-#include "Core/Functions.hpp"
-#include "Core/ParameterList.hpp"
-#include "Core/FitParameter.hpp"
-#include "Core/ParObserver.hpp"
+#include "Functions.hpp"
+#include "ParObserver.hpp"
+#include "ParameterList.hpp"
```
hansgans: In the header file the full path should be specified here

Core/Properties.cpp
```diff
@@ -75,49 +106,45 @@ void ReadParticles(PartList &list,
...
-    pt.add_child("QuantumNumber",tmp);
+    tmp.put("<xmlattr>.Class", "Int");
+    tmp.put("<xmlattr>.Type", i.first);
+    tmp.put("<xmlattr>.Value", i.second);
+    pt.add_child("QuantumNumber", tmp);
   }
-  pt.add_child("DecayInfo",DecayInfo);
+  pt.add_child("DecayInfo", DecayInfo);
   return pt;
+}*/
```
hansgans: Is it possible to save  Properties list with updated values?

Data/CorrectionTable.cpp
```diff
@@ -16,7 +16,7 @@ void CorrectionTable::Print() const {
   if (sys.size() == 0 && antiSys.size() == 0)
     return; // don't print if empty
   std::stringstream out;
-  ComPWA::TableFormater table(&out);
+  ComPWA::Tools::TableFormatter table(&out);
   table.addColumn("Bin", 12);             // add empty first column
```
hansgans: `TableFormatter` is located in `Core` but is in namespace `Tools`. That could be a bit more consistent.

Optimizer/Minuit2/MinuitIF.hpp
```diff
@@ -2,106 +2,62 @@
...
+
 namespace ComPWA {
 namespace Optimizer {
 namespace Minuit2 {
 
+struct OptimizationSettings {
+  bool UseMinos = false;
+  bool UseHesse = true;
+};
+
```
hansgans: I would prefer to simply add those as member variables

Optimizer/Minuit2/MinuitResult.cpp
```diff
@@ -2,275 +2,44 @@
...
+    os << "    *** COVARIANCE MATRIX NOT ACCURATE! ***\n";
   if (!CovPosDef)
-    out << "    *** COVARIANCE MATRIX NOT POSITIVE DEFINITE! ***" << std::endl;
+    os << "    *** COVARIANCE MATRIX NOT POSITIVE DEFINITE! ***\n";
   if (HesseFailed)
-    out << "    *** HESSE FAILED! ***" << std::endl;
-  if (HasValidCov) {
-    if (printCovMatrix) {
-      out << "COVARIANCE MATRIX:" << std::endl;
-      TableFormater *tableCov = new TableFormater(&out);
```
hansgans: Is the covariance not printed?

Tools/FitFractions.hpp
```diff
@@ -25,6 +26,17 @@
 namespace ComPWA {
 namespace Tools {
 
+struct FitFraction {
+  FitFraction(std::string name, double val, double err)
+      : Name(name), Value(val), Error(err) {}
```
hansgans: We could use `ComPWA::FitParameter` here and avoid another class

Core/FitParameter.hpp
```diff
@@ -1,230 +1,67 @@
...
-    ar &make_nvp("Fix", IsFixed);
-    ar &make_nvp("Value", Value);
-    ar &make_nvp("Bounds", Bounds);
-    try {
-      ar &make_nvp("ErrorType", ErrType);
-      ar &make_nvp("Error", Error);
-    } catch (...) {
-      Error = std::pair<double, double>(0, 0);
-      ErrType = ErrorType::SYM;
+template <typename T> struct FitParameter {
```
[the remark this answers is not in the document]
spflueger: Hmm, yeah I was just thinking for future options. You don't like this?

Core/FunctionTree/FunctionTree.cpp
spflueger: I just removed the full path of all includes from the same module, since that makes it a bit easier when modules alltogether

Core/FunctionTree/FunctionTreeEstimatorWrapper.hpp
spflueger: UserParameters save the state of the fit parameters when read from the xml (so default values as stated in the xml). Its a workaround to the following problem. The problem is, that I have to reset all fit parameters to NOT fixed, so that I satisfy the new Intensity/Estimator interface.

Core/FunctionTree/FunctionTreeIntensityWrapper.cpp
spflueger: Yeah there are better ways, but this was the easiest change for me to make it work. I would put improvements into this direction into the FuctionTree improvement project. Note that this has no influence on the fitting performance since this only occurs in the intensity and not the estimator

Core/FunctionTree/FunctionTreeIntensityWrapper.cpp
spflueger: Then only the needed first `N = Data.mDoubleValues().size()` entries are traken from the supplied data vector

Core/FunctionTree/ParameterList.hpp
spflueger: Whats the reasoning behind this?

Core/FunctionTree/TreeNode.hpp
spflueger: Hmm, I prefer includes from the same module to be without the full path. That makes more portable

Core/Properties.cpp
spflueger: No, I removed this functionality. It was too much work to keep this functionality and it is not necessary as it already exists elsewhere (analogous to the saving of a intensity). You can always reset the state of your model, by applying a fit result to a model.

Optimizer/Minuit2/MinuitIF.hpp
spflueger: Ok, I will make them members

Optimizer/Minuit2/MinuitResult.cpp
spflueger: The covariance matrix is now part of the general FitResult. So it will be printed if it is non empty

Tools/FitFractions.hpp
spflueger: Hmm I don't like this kind of reusing, since it will lead to a bad design eventually. A fit fraction is not a fit parameter in my opinion. Its just some value with an error that is calculated from a fit result (and an intensity). If there are some problems or downside we have to think about where in our current design the trouble arises

Core/FitParameter.hpp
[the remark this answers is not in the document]
hansgans: Well as I said, I can not imagin another type than double here. E.g. complex wouldn't make sense. If you prefer it we can keep it

Core/FunctionTree/FunctionTree.cpp
hansgans: Didn't you also change in install location? The reason why the full path as included in the header files is that after installation the path also needs to be correct.

Core/FunctionTree/FunctionTreeIntensityWrapper.cpp
hansgans: But the effect on generating data is huge, since also the normalization terms (which are constant after data has changed) are recalculated. Are you sure that with the current FunctionTree implementation is recalculation is not done properly?

Core/FunctionTree/FunctionTree.cpp
spflueger: I see. But inside a cpp file this would not matter. I thought the compiler will always find the header file if its in the same directory as the current header file (that includes the new header). And since these modules are always bundled together then there should be no problem (https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html)

hansgans: Build ComPWA is of course not a problem. But the header files are installed in a directory structure. And within this structure the path also needs to be correct. E.g. we have two header files which are installed in `$INSTALL_DIR`:
```
$INSTALL_DIR/include/ComPWA/Core/Parameter.hpp
$INSTALL_DIR/include/ComPWA/Core/ParameterList.hpp
```
If I now include `ParameterList.hpp` in another project `Parameter.hpp` needs to be found. I add the include path `include/ComPWA`. If we place `#include "Parameter.hpp"` in `ParameterList.hpp. This will not work.

Core/FunctionTree/FunctionTreeIntensityWrapper.cpp
spflueger: Thats true. I will try to do it correctly.

Core/FunctionTree/FunctionTree.cpp
spflueger: I thought that this does work... I will try it

spflueger: Yeah this works for me. I don't understand the reasoning behind this. But I just checked the gnu include files and they also specify the full path (at least where I looked). Well I guess there wont be any confusion, in case there would be multiple files with the same name. I will just change it to the full path.

Core/FunctionTree/FunctionTreeIntensityWrapper.cpp
spflueger: Ok just adding a `Notify()` in `Value::setValue()` analogous to `FitParameter::setValue()` worked! So this should be solved as well

spflueger:
Btw, I should mention that I commented out all of the fit fraction functionality, since this would require quite a bit of work. Since we want to change a bit more there, I would put these changes in a separate pull request.

spflueger:
Ok, I fixed the stuff you mentioned and fixed the other two things that were remaining (actually there was still a bug in the functiontree implementation of the WignerD). Now everything should compile and all test should run fine.

Core/FitParameter.hpp
```diff
@@ -1,230 +1,67 @@
...
+  bool IsFixed = false;
+  T Value;
+  std::pair<T, T> Error;
+  std::string Name = "";
+  std::pair<T, T> Bounds;
+
+  friend std::ostream &operator<<(std::ostream &os,
+                                  const FitParameter<double> &x) {
+    os << x.Name << ": " << x.Value;
+    if (x.Error.first != 0.0 || x.Error.second != 0.0) {
```
zhang-jingqing: Is this a good way to check if some double variable is equal to a specific value? Or it is a comparison using !=, so we can use it safely?

Core/FitParameter.hpp
```diff
@@ -1,230 +1,67 @@
...
+      }
     }
+    if (x.IsFixed)
+      os << " (fixed)";
+    return os;
   }
 };
-BOOST_SERIALIZATION_SHARED_PTR(ComPWA::FitParameter)
+
+using FitParameterList = std::vector<FitParameter<double>>;
```
zhang-jingqing: If we decide use a template for FitParameter, then I think here the FitParameterList a template will be better

Core/FitParameter.hpp
```diff
@@ -1,230 +1,67 @@
...
 
 } // namespace ComPWA
 
+namespace boost {
+namespace serialization {
+
+template <class Archive>
+void serialize(Archive &ar, ComPWA::FitParameter<double> &Par,
```
zhang-jingqing: If FitParameter is template class, then it will be better the serialize also support the template

Core/FitResult.cpp
```diff
@@ -4,232 +4,148 @@
...
-void FitResult::setFinalParameters(ParameterList &finPars) {
-  FinalParameters.DeepCopy(finPars);
-}
+namespace ComPWA {
 
-void FitResult::setTrueParameters(ParameterList &truePars) {
-  TrueParameters.DeepCopy(truePars);
+std::string makeFitParameterString(ComPWA::FitParameter<double> p) {
+  std::stringstream ss;
+  ss << std::setprecision(10);
```
zhang-jingqing: To cout full precision of a variable, we can do it like this:
https://stackoverflow.com/questions/554063/how-do-i-print-a-double-value-with-full-precision-using-cout
double d = 100.0000000000005;
int precision = std::numeric_limits<double>::max_digits10;
std::cout << std::setprecision(precision) << d << std::endl;
I think this will be better than a fixed precision(but maybe the printed number is very long...)

Core/FitParameter.hpp
spflueger: Hmm. Yeah, since I simply want to know if errors exist, the best would be to use `std::optional` ( which comes with c++17). I can also introduce a bool just like for the bounds. But sooner or later we should upgrade the bound and error to std::optionals.

Core/FitResult.cpp
spflueger: Do we need/want full precision there?

Core/FitParameter.hpp
spflueger: Not sure I can follow. Can you make an example?

spflueger:
Can I close this pull request, or are there still a few open topics to discuss?

hansgans:
I'm fine with it

PR #202, opened 2019-07-29
Spin improvements

spflueger:
This is a proposal to improve the Spin class

### DIFF
--- a/Core/CMakeLists.txt
+++ b/Core/CMakeLists.txt
@@ -47,11 +47,10 @@ install(TARGETS Core
 #
 # TESTING
 #
-# Testing routines are stored in separate directory
-#file(GLOB TEST_SRCS RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} test/*.cpp)
 set(TEST_SRCS
   test/CX0TestApp.cpp
   test/ParticleTest.cpp
+  test/SpinTest.cpp
 )
 
 #Run through each source
--- a/Core/Properties.cpp
+++ b/Core/Properties.cpp
@@ -17,14 +17,14 @@ ParticleProperties::ParticleProperties(boost::property_tree::ptree pt) {
 
       // We have to distinguish between spin and integer quantum numbers
       if (v.second.get<std::string>("<xmlattr>.Class") == "Spin") {
-        auto value = v.second.get<double>("<xmlattr>.Value");
-        double valueZ = 0.0;
+        double Mag = v.second.get<double>("<xmlattr>.Value");
+        Spin J(Mag);
         try { // Projection of spin is optional (e.g. (I,I3))
-          valueZ = v.second.get<double>("<xmlattr>.Projection");
+          double Proj = v.second.get<double>("<xmlattr>.Projection");
+          J = Spin(Mag, Proj);
         } catch (std::exception &ex) {
         }
-        spinQuantumNumbers_.insert(
-            std::make_pair(type, ComPWA::Spin(value, valueZ)));
+        spinQuantumNumbers_.insert(std::make_pair(type, J));
       } else if (v.second.get<std::string>("<xmlattr>.Class") == "Int") {
         auto value = v.second.get<int>("<xmlattr>.Value");
         intQuantumNumbers_.insert(std::make_pair(type, value));
--- a/Core/Properties.hpp
+++ b/Core/Properties.hpp
@@ -29,8 +29,6 @@ class ParticleProperties {
 public:
   ParticleProperties(boost::property_tree::ptree pt);
 
-  // boost::property_tree::ptree save();
-
   std::string getName() const { return Name; }
 
   pid getId() const { return Id; }
--- a/Core/Spin.hpp
+++ b/Core/Spin.hpp
@@ -2,26 +2,83 @@
 // This file is part of the ComPWA framework, check
 // https://github.com/ComPWA/ComPWA/license.txt for details.
 
-///
-/// \file
-/// Spin class.
-///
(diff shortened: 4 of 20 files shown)


hansgans:
Probably I miss something, but I do not understand what the principal problem in having a Spin class which represents interger of half-integer spins.

spflueger:
I don't quite understand your question. The Spin class remains, and still represents integral and half-integral spins. The problem is that a spin defines a Spin a magnitude and a spin projection, but your code mostly uses only one of these two separately. For example, a WignerD function takes the parameters `J`, `mu'`, `mu`, which are a spin magnitude and two spin projections. Currently three Spins are created (all ignoring the Spin projection) and passed around the Functions. This works because the Numerator of the Spin is allowed to be negative (int).
I introduced a new Fraction class to solve this problem. This might not be the best solution though. Another option would be to just pass around doubles and create a fraction from it when necessary (like casting it to int value).

Does this answer your question?

hansgans:
I understand the problem now. But I think it is very misleading to introduce a class `Fraction` which can also be an integer. I'm fine with keeping it as simple as possible and just pass double values for half-integer spins. And if I understand your implementation correctly, you construct a spin with a fixed projection. But the projection depends on the axis and can have several values [-J,...,J] (I'm not sure if this is relevant in this case).

spflueger:
>  But I think it is very misleading to introduce a class Fraction which can also be an integer.

I can't follow here. The reason why we introduce this in the first place, is not to describe the spin properties with `double`'s. Then code which work with the spin properties, can extract this information precisely and without problems (e.g. the WignerD and Clebsch Gordan functions). Since we can have integer and half-integer spins, but we need a common description for this, only fractions work here (the same counts for the spin class).

> And if I understand your implementation correctly, you construct a spin with a fixed projection. But the projection depends on the axis and can have several values [-J,...,J] (I'm not sure if this is relevant in this case).

You choose a specific axis and then the spin projection stays fixed. The Isospin of a particle also has a magnitude and a projection/z-component, both together define the full information. That's why I defined a spin as an object with a magnitude and a projection. Since we do not use the Isospin in any amplitude or intensity code, the spin class (including the projection) is not really needed by ComPWA. But maybe in the future, something like this might be necessary.

So what would be your solution? If we pass half-integer spins as doubles, then we should do that also for the integer spins. Otherwise we would have two different code versions, depending on type of spin... (one with double arguments and one with integers?)

spflueger:
I'm also not fully happy with the solution. But I can't come up with anything better so far. That's why I put it as a proposal so we can discuss it here. So I'm grateful for your comments @hansgans 

Do you have any opinions @zhang-jingqing ?

hansgans:
> > But I think it is very misleading to introduce a class Fraction which can also be an integer.
> 
> I can't follow here. The reason why we introduce this in the first place, is not to describe the spin properties with `double`'s. Then code which work with the spin properties, can extract this information precisely and without problems (e.g. the WignerD and Clebsch Gordan functions). Since we can have integer and half-integer spins, but we need a common description for this, only fractions work here (the same counts for the spin class).

My point is that you construct a spin object e.g. `ComPWA::Spin s(1)`. I find it misleading that `Spin::getMagnitude()` returns a `Fraction` in that case. 
> 
> > And if I understand your implementation correctly, you construct a spin with a fixed projection. But the projection depends on the axis and can have several values [-J,...,J] (I'm not sure if this is relevant in this case).
> 
> You choose a specific axis and then the spin projection stays fixed. The Isospin of a particle also has a magnitude and a projection/z-component, both together define the full information. That's why I defined a spin as an object with a magnitude and a projection. Since we do not use the Isospin in any amplitude or intensity code, the spin class (including the projection) is not really needed by ComPWA. But maybe in the future, something like this might be necessary.

I think the isospin z-component is not a good example here since it is not really a spin projection but rather a quantum number. The project of the spin of a particle depends on the axis on to which you project it. So I would not consider it to be fixed. Actually, I would remove the projection stuff from  the `ComPWA::Spin` class.
> 
> So what would be your solution? If we pass half-integer spins as doubles, then we should do that also for the integer spins. Otherwise we would have two different code versions, depending on type of spin... (one with double arguments and one with integers?)

I would keep the `Spin` class but pass doubles to `AmpFlatte` and `WignerD`.

spflueger:
> Actually, I would remove the projection stuff from the `ComPWA::Spin` class.

> I would keep the `Spin` class but pass doubles to `AmpFlatte` and `WignerD`.

Why do we need the Spin class in the first place then? We can just use doubles. I see no point in the Spin class then.
Right now we convert a double value, to a spin (which is a fraction). Then we convert it back to a double, and eventually into an int. This whole type conversion you usually try to avoid by constructing a class that accurately represents the objects.

The only way a Spin class makes sense to me is, that you can work with the precise (fraction information) of the spin and use it where ever it is needed (WignerD etc). So then we should not pass doubles to these needed places. So convert a double into a spin (maybe not even have a double in the first place). And that's it.

hansgans:
I agree to remove the Spin class completely. Than we should treat the spin of a particle as orinary quantum number in ParticleProperties (e.g. remove `ParticleProperties::getSpinQuantumNumber`).

spflueger:
> I agree to remove the Spin class completely. Than we should treat the spin of a particle as orinary quantum number in ParticleProperties (e.g. remove ParticleProperties::getSpinQuantumNumber).

Ok, you mean to use map<std::string, double> to store the spin?

spflueger:
So I would close this pull request. The conclusion was to abolish the Spin class and use doubles instead. Feel free to reopen if you think this topic needs more discussion

PR #204, opened 2019-08-02
Cleanup of old Intensity and Amplitude classes

### LINKED ISSUE #191: Cleanup of the current FunctionTree backend
** This group of tasks can only be done after the new interface design change **

* Remove obsolete code, more precisely all Amplitudes + Intensities (only the FunctionTree creation parts stay)
* Move Parameter handling code to the FunctionTree or Wrapper class
* Adjust IntensityBuilderXML, to only generate FunctionTrees instead of the old Intensity Objects
* Adjust the wrapper class to work with the new IntensityBuilderXML

spflueger:
- Closes #191
- Removed old Intensity and Amplitude classes.
- Refactored IntensityBuilderXML to build FunctionTrees and
  FunctionTreeIntensityWrapper instead.
- General versions of the Estimators are fully implemented. Here the
  Intensity instance is evaluated once with the Data sample and once
  with the Phase space sample. This as the disadvantage that no caching
  can be performed.
- Fixed bug in some dynamic functions, where a constant mass parameter
  was used for the decay products.
  (Flatte Couplings are still constants! This remains to be fixed.)
- Currently the creation of a full FT Estimator only works within the
  builder class. It would be convenient to improve the builder, to allow
  injections from outside the class.
- Creation of full SumLogLH Estimator has still be be implemented.


spflueger:
This is just the first version of the cleanup of the old Intensity and Amplitude classes. You can check it out and get an idea of how its supposed to be.

One open issue is the way the Estimators are created. Currently I create the FunctionTree version of MinLogLH from a IntensityWrapper instance and the new clone functionality of the IntensityWrapper. Let me know what you think of this. I don't really like the solution, because I have to give public access to a lot of member functions of the IntensityWrapper.

Maybe it is better to just create the Estimator through the IntensityBuilderXML. Maybe you have some better ideas.

Examples/DalitzAnalysis_D0KsKK/DalitzFitD03K.cpp
```diff
@@ -24,6 +24,7 @@
 // Root header files go here
 #include "TFile.h"
 
+#include "../../Physics/HelicityFormalism/WignerD.hpp"
```
weidenka: Shouldn't that be an absolute path?

spflueger: That Example is currently taken out of the build and was not adjusted for quite a while now. I did not want to spend too much time adjusting all of these examples, until the cleanup is completely finished.
I'm also not sure if we can remove some of the examples completely.

Physics/Dynamics/FormFactor.hpp
```diff
@@ -120,6 +122,12 @@ inline double FormFactor(std::complex<double> qValue, unsigned int orbitL,
     // Reference: S.U.Chung Annalen der Physik 4(1995) 404-430
     // z = q / (interaction range). For the interaction range we assume
     // 1/mesonRadius
+
+    // TODO: is this comment actually accurate?
```
weidenka: I think you put that here

Physics/Dynamics/FormFactorDecorator.hpp
```diff
@@ -19,62 +14,32 @@ namespace ComPWA {
...
 namespace Dynamics {
 
-/// The production formfactor is implemented in Blatt-Weisskopf type
-/// \f$B_{L}(q)\f$ Ref. Dalitz Plot Analysis, PDG 2012, where \f[
-/// \frac{B^{\prime}(q_{0})}{B^{\prime}_{L}(q)} = q^{L}B^{\prime}_{L}(q, q_{0})
-/// \f]
-/// \f$B_{L}(q)(i.e., F_L(q))\f$ with \f$L\f$ up to 4 can be found in
-/// BNL-QGS-06-101
-class FormFactorDecorator : public AbstractDynamicalFunction {
+namespace ProductionFormFactor {
```
weidenka: Why is it called ProductionFormFactor? It should be the same for a particle decay.

Physics/HelicityFormalism/HelicityKinematics.hpp
```diff
@@ -75,14 +75,6 @@ class HelicityKinematics : public ComPWA::Kinematics {
       const std::vector<pid> &finalState,
       const ComPWA::FourMomentum &cmsP4 = ComPWA::FourMomentum(0, 0, 0, 0));
 
-  /// Delete copy constructor. For each Kinematics in the analysis only
-  /// one instance should exist since Kinematics does the bookkeeping which
-  /// SubSystems variables are needs to be calculated. That instance can then be
-  /// passed as (smart) pointer.
-  /// Note1: Not sure if we also should delete the move constructor.
-  /// Note2: We have to delete the copy constructor in Base and Derived classes.
-  HelicityKinematics(const HelicityKinematics &that) = delete;
```
weidenka: Isn't that statement still true? Why should we not delete the copy constructor here?

Physics/IntensityBuilderXML.hpp
```diff
@@ -19,48 +20,51 @@
...
-
-  std::shared_ptr<NamedAmplitude>
-  createAmplitude(std::shared_ptr<PartList> partL,
-                  std::shared_ptr<Kinematics> kin,
-                  const boost::property_tree::ptree &pt) const;
+                           const boost::property_tree::ptree &pt);
+
+  std::pair<ComPWA::FunctionTree::FunctionTreeEstimatorWrapper,
+            FitParameterList>
+  createMinLogLHFunctionTreeEstimator(
```
weidenka: Shouldn't that function be part of MinLogLH?

spflueger: Yeah, thats the part I'm still working on/thinking about. The problem is that you need to make a copy of the Intensity FunctionTree for the MinLogLH Estimator. Because you need to evaluate once the actual data sample and once the phsp sample. Since only the Intensity Builder has the FunctionTree information (which is currently private, since the building functions return the FunctionTrees are private), the creation of the MinLogLH Estimator has to go inside this class.
I dont really like this either. Maybe we have to make the FunctionTree building function public. Or do you have any ideas?

Physics/HelicityFormalism/HelicityKinematics.hpp
spflueger: Hmm, yeah that is true. I have to double check why I removed this....

Physics/Dynamics/FormFactor.hpp
spflueger: Ups, don't know how that ended up there... This comment comes from FormFactorDecorator.hpp

Physics/Dynamics/FormFactorDecorator.hpp
spflueger: Good question. Yeah I can make it a general FormFactor. I would move all of the code to the FormFactor.hpp and get rid of the FormFactorDecorator then.

Physics/IntensityBuilderXML.hpp
weidenka: We could just build two Intensities, one for the data sample and one for the phase space sample. I mean if you have one tree which is one evaluated with the data and and afterwards with the phase space sample, than you basically don't cache anything, since the tree is recalculated at each change of the data sample

spflueger: Yeah that's what I'm doing right now already. Its just a matter of code design.

Physics/Dynamics/FormFactorDecorator.hpp
spflueger: Moved relevant code to FormFactor.hpp/cpp and removed the decorator ones

Examples/DalitzAnalysis_D0KsKK/DalitzFitD03K.cpp
spflueger: I would shift this to later when everything is fixed. Its a bit tedious to always keep all of the examples up to date.

weidenka: ok

Physics/Dynamics/Voigtian.hpp
```diff
@@ -42,106 +35,74 @@ namespace Dynamics {
...
+  double argu = sqrtS - mR;
+  double c = 1.0 / (sqrt(2.0) * sigma);
+  double a = c * 0.5 * wR;
+  double u = c * argu;
+  std::complex<double> z(u, a);
+  std::complex<double> v = Faddeeva::w(z, 1e-13);
+  double val = c * 1.0 / sqrt(M_PI) * v.real();
+  double sqrtVal = sqrt(val);
+
+  /// keep the phi angle of the complex BW
```
zhang-jingqing: Now I think this thought (keep the phi angle of the complex BW) is too naive and meaningless, and it may also cause difficulty to understand what the voigtian is done. So I think it is better to just return a real number for this voigtian function: std::complex<double> result(sqrtVal, 0)

spflueger: Hmm, why can you remove the phase?

Physics/Dynamics/RelativisticBreitWigner.hpp
```diff
@@ -5,149 +5,124 @@
...
+  // Phase space factors at sqrt(s) and at the resonance position
+  auto phspFactorSqrtS = phspFactor(sqrtS, ma, mb);
+  auto phspFactormR = phspFactor(mR, ma, mb);
+
+  // Check if we have an event which is exactly at the phase space boundary
+  if (phspFactorSqrtS == std::complex<double>(0, 0))
+    return std::complex<double>(0, 0);
+
+  std::complex<double> qRatio =
+      (phspFactorSqrtS / phspFactormR) * mR / sqrtS;
```
zhang-jingqing: i think the equation qRatio should be : qRatio = pshpFactorSqrtS/phspFactormR, no * mR/sqrts, since pshpFactor \propto 2q/sqrtS

spflueger: Ah thats true. So just to be sure I remove the whole mR/sqrtS factor, right?

spflueger: I changed it now. Can you double check and resolve it if its correct?

weidenka: I think the original code was correct since `phspFactormR` is the phsp factor evaluated at the mass of the resonance - so mR and sqrtS do not cancel.

spflueger: The way I see it: We want to calculate `q_sqrtS/q_mR * mR/sqrtS`. That is the factor that goes in front of the Barrier Factors B_L. Since `q = phspFactor * sqrtS` (analogous for mR), `q_sqrtS/q_mR * mR/sqrtS  = phspFactorSqrtS/phspFactormR`

Physics/Dynamics/Voigtian.hpp
zhang-jingqing: We can't. But the true phase from the Breit-Wigner is not available due to the resolution. And it is not clear how to convolute a complex function with a resolution function. In the old code, I assume that the smeared function (e.g., |BW|^2oxGaus) as a complex function has the same phase with the original BW. Usually this is not true (even as an approximation it is not true in some cases, e.g., pi0/eta/D has a width at keV level, resolution at MeV level). So I think remove this assumption is better.

spflueger: Puh, ok. Should we make a separate issue + pull request for that?

zhang-jingqing: Yes, another issue and pull request will be more clear.

Physics/IntensityBuilderXML.hpp
spflueger: I just had an idea. The comment you made above would also be solved this way. What if we adjust the FunctionTree backend this way that on a copy operation an actual copy will be made. Usually the move operation will do just fine. But in case of creating a estimator making a copy would be needed. I think that makes things really transparent and easy. What do you think?

Physics/Dynamics/Voigtian.hpp
spflueger: Ok, I leave that to you :)

Physics/Dynamics/RelativisticBreitWigner.hpp
weidenka: No we want to calculate `q_sqrtS/q_mR` and we use the `rho_sqrtS` and `rho_mR` to do so. (for `rho` the analytic continuation is properly implemented).

spflueger: I disagree. I got my information from here: http://pdg.lbl.gov/2019/reviews/rpp2018-rev-resonances.pdf, equation 48.23. The power factor of `2L+1` will be 1 in our case, due to the different definitions of the form factors `F_L`.

Physics/IntensityBuilderXML.hpp
weidenka: I'm not sure if I understand what you mean. Why is a copy mandatory for an estimator?

spflueger: What I meant with that is, that for the construction of the MinLogLH you would need to copy your Intensity once and evaluate that copy with the phsp sample instead.
If we use the approach with the copy operator/constructor, we have the advantage to only make a single implementation for a concrete estimator (ie MinLogLH). That implementation could be used with any Intensity no matter what actually performs the calculation underneath (functiontree etc)

Physics/Dynamics/RelativisticBreitWigner.hpp
weidenka: Ok I see what you mean. But the term `qRatio^(2L+1)` is meant to be `(q_sqrtS/q_mR)^(2L+1)` so the code above is correct. The additional factor `mR/SqrtS` is not exponentiated. And was added to the describtion in PDG2018 or so. Given the PDG2014 reference it is correct but if course we should update it.

Physics/IntensityBuilderXML.hpp
weidenka: We have to consider that the FunctionTree can become very memory consuming (depending on the amplitude model). I think we should try to avoid a "deep" copy. Probably we could delete the copy constructor to ensure that only a move is possible.

Physics/Dynamics/RelativisticBreitWigner.hpp
spflueger: Afaik this `2L+1` is the mistake that was pointed out by @zhang-jingqing . Since his corrections are not merged yet. I would have just done that. The main problem is, that we use the formula from the pdg, but not the form factors of the pdg. Thats where these "other" equations in our code stem from

Physics/IntensityBuilderXML.hpp
spflueger: Well for the construction of the MinLogLH estimator, a copy of the function tree is inevitable. No matter what you do, one copy has to be made. I agree that a "deep" copy operation could lead to misuse. Although I don't see it as a problem because:
- Unless you write bad code, the object would not be copied (but moved or reference)
- Its natural to assume, that a copy is a "deep" copy. This is just the c++ way
For me the pros outweigh the cons here. Which solution do you recommend instead?

Physics/Dynamics/RelativisticBreitWigner.hpp
weidenka: What is the difference between the form factors we use and those suggested by the PDG?

Physics/IntensityBuilderXML.hpp
weidenka: The full FunctionTree should not reside in memory twice. I think thats very important.

Physics/Dynamics/RelativisticBreitWigner.hpp
zhang-jingqing: The form factors we used are same to the B_L(q) in DalitzAnalysis, PDG http://pdg.lbl.gov/2012/reviews/rpp2012-rev-dalitz-analysis-formalism.pdf
There is also another type B'_L(q) in same reference, and they have the relation: 
B_L(q)/B_L(q_0) = (q/q_0)^L\times B'_L(q, q_0).
The F_L(q) used in Resonance in PDG is the third form., I think after some transformation 
F_L(q) can be related to B_L(q) by: q^L\times F_L(q) = B_L(q).
In SU Chung's references, he sometimes also use F_L(q) to denote the form factor, but is different to the definition used in PDG Resonance chapter.

Physics/IntensityBuilderXML.hpp
spflueger: Hmm you mean it should be impossible (deleting copy constructor)? If I imagine everything correctly, our code will not generated it twice. However the user might create a copy of it...
If we want to delete the copy constructor then we have to search for another solution.

The large memory consumption is also a problem of the function tree right now, since everything is cached. I think if we implement some wiser caching, the consumption will be much lower, without losing any speed (maybe even faster).

weidenka: The current implementation - as far as I understand it - allows to copy the FunctionTree via the standart copy operator i.e. the shared_ptr's for two copies point to the same objects. Thats fine for me, but one could say it is a bit misleading since - as you said - one would expect that a 'deep' copy is done. 
* I would disfavour the solution in which a 'deep' copy is implemented since it is possible to create the tree several times in memory and I do not see a nessesity to have this functionality.
* Deleting the copy constructor would make clear that the tree should never be copied.
But as I said the current code is fine for me

spflueger: Hmm ok, then I have to come up with something better. The current solution is not very flexible for the creation of the estimators, because right now they have to be part of the `Builder`. Currently I'm thinking about your idea (@weidenka) to normalize every intensity automatically on creation. Then the estimator would only have to evaluate the Intensity and thats it. Also the estimators would not need to know about the underlying implementation of Intensity. Seems perfect! I can't think of any downside of this so far? Can you guys? Maybe this is the best solution for the problem.
I will give this a try next week, and hopefully I can wrap up this pull request then.

weidenka: That sounds good to me

spflueger: > Also the estimators would not need to know about the underlying implementation of Intensity.

Actually this statement is not true. In that way the FunctionTree will not perform caching for the data evaluation. In order to get the full caching feature, a full FunctionTree estimator has to be created. That only works by having the information of the intensity FunctionTree in the first place. This also makes the creation process of the estimators equally difficult (similar problems as the other solutions).

Physics/Dynamics/RelativisticBreitWigner.hpp
spflueger: So is the current implementation ok now? If so can I resolve this part?

Physics/IntensityBuilderXML.hpp
spflueger: So this part is also "finished" now. The outermost intensity is automatically normalized (even if the no NormalizedIntensity tag is used). I still had to give the `FunctionTreeEstimatorWrapper` a public access to the FunctionTree members, which is not good. But I don't see any way around this right now, except what I described above.

spflueger:
I should mention that I commented out the FormFactorDectorator tests, since this code is not compatible anymore (and requires a bit more effort). Either we drop them completely or they need a rework. I would open a separate issue for that

Physics/Dynamics/RelativisticBreitWigner.hpp
weidenka: We should keep the current implementation at least until this large pull request is merged. The would make debugging easiert. And afterwards handle changes to the form factor description in a separate pull request.

spflueger:
Ok, the pull request if finished from my side. You can check it again. If you are happy or not happy with it let me know. If you guys are satisfied with it then i will merge it

PR #207, opened 2019-09-02
Removed Spin class

### LINKED ISSUE #199: Spin SetDenominator in operator+ or operator-
By adding/subtracting spins it might happen that the denominator is not equal to 1 or 2.
This results that SetDenominator throws an error.

At this point SetDenominator shoundn´t throw the error, since simplify later on reduces the denominator again.

spflueger:
- Fixes #199
- Removed Spin class. Instead double values are used. The orbital
  angular momentum is declared as a unsigned int.
- Improved the interface of the Properties class. Now the quantum
  numbers are retrieved via a template member function. Later on one
  might think of upgrading the std::maps storing the quantum numbers to
  a std::variant.
- Commented out recursive Wigner_d calculations, which relied on the
  Spin class and would need a rework. Since they are not used by and
  code, removing them seems fine.


weidenka:
Fine to me. Could to use //-style comments instead of /* ... */? // is a bit easier to read in diffs.

PR #209, opened 2019-09-06
Fixed bug in the relativistic Breit-Wigner

spflueger:
- The barrier term in the denominator of the relativistic BW now
  correctly resembles the definition. The incorrect implementation
  stemmed from the use of different form factors and the phase space
  factor instead of the breakup momentum, without modifing the equation
  accordingly.

### DIFF
--- a/Physics/Dynamics/RelativisticBreitWigner.hpp
+++ b/Physics/Dynamics/RelativisticBreitWigner.hpp
@@ -36,16 +36,17 @@ struct InputInfo {
 
 ///
 /// Relativistic Breit-Wigner model with barrier factors.
-/// The dynamical function implemented here is taken from PDG2014 (Eq.47.22)
+/// The dynamical function implemented here is taken from PDG2018 (Eq.48.22)
 /// for the one channel case. The dynamic reaction
 /// \f[
 /// \mathcal{A}_R(s) = \frac{g_p*g}{s - M_R^2 + i \sqrt{s} \Gamma_R B^2}
 /// \f]
 /// \f$ g_p, g\f$ are the coupling constants for production and decay and
-/// the barrier term \f$ B^2\f$ is parameterized according to Eq.47.23:
+/// the barrier term \f$ B^2\f$ is parameterized according to Eq.48.23:
 /// \f[
-///     B^2 = \left( \frac{q(s)}{q(s_R)} \right)^{2L+1} \times
-//                     \left( \frac{F(s)}{F(s_R)} \right)^{2}
+///     B^2 = \left( \frac{q(\sqrt{s})}{q(M_R)} \right)^{2L+1} \times
+///           \left( \frac{M_R}{\sqrt{s}} \right) \times
+///           \left( \frac{F(\sqrt{s})}{F(\sqrt{s_R})} \right)^{2}
 /// \f]
 /// This corresponds to the Blatt Weisskopf form factors B_L like
 /// \f[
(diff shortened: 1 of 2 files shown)


Physics/Dynamics/RelativisticBreitWigner.hpp
```diff
@@ -78,14 +79,17 @@ dynamicalFunction(double mSq, double mR, double ma, double mb, double width,
   if (phspFactorSqrtS == std::complex<double>(0, 0))
     return std::complex<double>(0, 0);
 
-  std::complex<double> qRatio =
-      std::pow((phspFactorSqrtS / phspFactormR) * mR / sqrtS, (2 * L + 1));
+  // The each FormFactor includes a term q^L. Therefore only q instead of
+  // q^(2L+1) has to be calculated. Also because phspFactor ~ q/sqrt(s) (see
+  // PDG2018, equation 48.2) the factor \frac{M_R}{\sqrt{s}} can also be omitted
```
weidenka: This is correct to my opinion: `qRatio = qSqrtS / qmR = phspFactorSqrtS / phspFactormR) * sqrtS / mR` 
I'm not sure about the exponent. The implemented form factor is `B_L = q F_L`. Than you still need `qRatio^(2L)`.

spflueger: Equation 48.23 in http://pdg.lbl.gov/2019/reviews/rpp2018-rev-resonances.pdf is what we want to calculate. Because our form factors are different, more precisely `B_L = q^L * F_L` the exponent `2L+1` in equation 48.2, reduces to just `1`.
Now from equation 48.2 we know `rho ~ q / sqrt(s)`. So the term `q_c/q_R * M_R/sqrt(s) = rho_c/rho_R`

weidenka: To my understanding the difference is `B_L = q * F_L`.

spflueger: Can you verify your calculation and post it here if it still does not match

weidenka: I compare `F_L` given in http://pdg.lbl.gov/2019/reviews/rpp2018-rev-resonances.pdf Section 48.2 and the formula implemented in `FormFactor.cpp`.

spflueger: But you see that the pdg `F_L` do not have any q terms in the nominator, while our FormFactor implementation does

zhang-jingqing: > To my understanding the difference is `B_L^2 = q * F_L^2`.

Hi, as I understand the F_L in rpp20180rev-resonances and our B_L is related by q^LF_L = B_L
It should be easier find the BW and formfactor formula in this: http://pdg.lbl.gov/2012/reviews/rpp2012-rev-dalitz-analysis-formalism.pdf
Table 2 and Eq. (7) and Eq. (8). I think this kind of BW should be equal to that in rpp-2018-rev-resonances

weidenka: Ah ok I see my mistake. I agree on B_L but still the `qRatio = phspFactorSqrtS / phspFactormR` is not correct. Furthermore, we need to think how the propagate the analytic continuation on phspFactor to `F_L` or `B_L`.

spflueger: Hmm yeah. It seems almost to me, that the K-matrix phase space factor definitions can't be used like this in the BW form. It would be best to use equation 48.24 for the analytic continuation of q. Then the phase space factor rho should not be used in the BW and Flatte anymore.

spflueger: Or can we keep rho as it is except that it does not use the q anymore, but uses this rho hat definition. Then we simply define q ~ rho * sqrt(s)? But here we would have to use a complex value for s < 0 [so sqrt(s) = i * sqrt(|s|) for s < 0]

spflueger:
>I was rethinking my argument: At threshold L=0, so you anyway only have the term q/qR = rho/rho_R sqrtS/mR and the formfactors are one. So we do not need to propagate the proper calculation of q to the form factors.

And how about L > 0?

weidenka:
At threshold you can only have L=0, I think.

PR #210, opened 2019-09-09
Small improvement of the Function interface

spflueger:
- Introduced new struct Parameter, that consists of a value and a name.
- The `getParameters()` function now returns a vector of Parameters,
  instead of doubles. This allows the identification of parameters and
  their ordering.

### DIFF
--- a/Estimator/test/MinLogLHEstimatorTest.cpp
+++ b/Estimator/test/MinLogLHEstimatorTest.cpp
@@ -22,29 +22,37 @@
 class Gaussian : public ComPWA::Intensity {
 public:
   Gaussian(double mean, double width)
-      : Mean(mean), Width(width), Strength(1.0) {}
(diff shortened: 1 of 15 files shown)


Estimator/test/MinLogLHEstimatorTest.cpp
```diff
@@ -22,29 +22,33 @@
...
+                                       std::pow(Width.Value, 2));
     });
     return result;
   }
 
   void updateParametersFrom(const std::vector<double> &params) {
-    Mean = params[0];
-    Width = params[1];
-    Strength = params[2];
+    Mean.Value = params[0];
```
weidenka: You might check the length of `params` here.

spflueger: I added a check here. Just a general remark: a simple length check is usually not necessary. I personally would allow for example also smaller `vector<double>`.

Issue #206, opened 2019-08-30
Calculate error of MC integration


weidenka:

spflueger:
Just for reference https://en.wikipedia.org/wiki/Monte_Carlo_integration

weidenka:
I implemented the calculation of the uncertainty. For one example model the uncertainty is smaller than the machine precision for 10^6 phsp events. But if I choose a different seed for the generation of the phsp sample the deviation is notably. Currently, I don't understand where is coming from.

PR #211, opened 2019-09-09
Add error calculation to MC integration

weidenka:
Add error calculation and unit test of MC integration, see Issue #206


Data/DataSet.hpp
```diff
@@ -23,6 +23,13 @@ struct DataSet {
   std::vector<std::string> VariableNames;
 };
 
+inline void resize(DataSet &set, int size) {
```
spflueger: use `unsigned int` or `size_t`

Data/Root/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ add_library(RootData
   SHARED ${lib_srcs} ${lib_headers}
 )
 
+target_include_directories(RootData PUBLIC ${ROOT_INCLUDE_DIR})
```
spflueger: What is that for? How did the headers get found before?

Tools/test/IntegrationTest.cpp
```diff
@@ -0,0 +1,249 @@
+// Copyright (c) 2013, 2017 The ComPWA Team.
+// This file is part of the ComPWA framework, check
+// https://github.com/ComPWA/ComPWA/license.txt for details.
+
+#define BOOST_TEST_MODULE SaveAndLoadFitResultTest
```
spflueger: Probably rename this?

Tools/test/IntegrationTest.cpp
```diff
@@ -0,0 +1,249 @@
...
+#include "Data/DataSet.hpp"
+#include "Data/Generate.hpp"
+#include "Data/Root/RootGenerator.hpp"
+#include "Estimator/MinLogLH/MinLogLH.hpp"
+#include "Optimizer/Minuit2/MinuitIF.hpp"
+#include "Physics/HelicityFormalism/HelicityKinematics.hpp"
+#include "Physics/IntensityBuilderXML.hpp"
+#include "Tools/UpdatePTreeParameter.hpp"
+#include "Tools/Integration.hpp"
+
```
spflueger: You probably don't need all of those headers. Also the boost ones

Tools/test/IntegrationTest.cpp
```diff
@@ -0,0 +1,249 @@
...
+  // Kinematics
+  ModelStream.clear();
+  ModelTree = boost::property_tree::ptree();
+  ModelStream << JpsiDecayKinematics;
+  boost::property_tree::xml_parser::read_xml(ModelStream, ModelTree);
+  auto DecayKin = Builder.createHelicityKinematics(
+      PartL, ModelTree.get_child("HelicityKinematics"));
+
+  // Estimated using 10^7 events: 0.198984549+-5.627720404e-05
+  double TrueIntegral = 0.198984549;
```
spflueger: How did you calculate the true integral?

Tools/Integration.cpp
```diff
@@ -17,20 +17,60 @@
 namespace ComPWA {
 namespace Tools {
 
-double integrate(ComPWA::Intensity &intensity,
+struct KahanSummation {
+  double sum;
+  double correction;
+  operator double() const { return sum; }
```
spflueger: I would not use the cast, but instead return this KahanSummation from the accumulate.

Tools/test/IntegrationTest.cpp
weidenka: Its estimated using a high number of phsp events. I hardcode it here to save memory on TravisCI and also to indicate if the model changes

Tools/Integration.cpp
weidenka: Is there a disadvantage?

Tools/test/IntegrationTest.cpp
spflueger: Ok, thats fine. I think it would be not bad if we would also include a test for the correctness of the integration (accuracy test). So maybe integrate a gaussian or a sphere.

Tools/Integration.cpp
spflueger: I think its harder to understand than without. At least I asked myself what that is for and had to look through the code

weidenka: After summation you use it as a `double` so a cast seems fine to me. To understand what the code does it is not necessary to understand the details of KahanSummation. I prefer it this way.

Issue #160, opened 2019-03-29
Coverity scan errors


weidenka:
I pushed the recent master branch to coverity scan (see that batch in the README). There is a long list of potential errors. Someone should carefully go through it.

spflueger:
So I guess this issue is just partly resolved with that pull request?

PR #212, opened 2019-09-10
Fix issues raised by CoverityScan

weidenka:
Issue #160

### DIFF
--- a/Examples/SimFit/SimFit.cpp
+++ b/Examples/SimFit/SimFit.cpp
@@ -217,7 +217,6 @@ energyPar createEnergyPar(size_t NEvents, double SqrtS, int seed,
 
   auto data =
       ComPWA::Data::generate(NEvents, kin, RandomGenerator, amp, mcSample);
-//  return energyPar(NEvents, std::move(kin), amp, data, mcSample);
   return energyPar{NEvents, std::move(kin), amp, data, mcSample};
 }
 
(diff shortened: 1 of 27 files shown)


Examples/SimFit/SimFit.cpp
```diff
@@ -224,7 +217,8 @@ energyPar createEnergyPar(size_t NEvents, double SqrtS, int seed,
 
   auto data =
       ComPWA::Data::generate(NEvents, kin, RandomGenerator, amp, mcSample);
-  return energyPar(NEvents, std::move(kin), amp, data, mcSample);
+//  return energyPar(NEvents, std::move(kin), amp, data, mcSample);
```
spflueger: Could you remove the comment. I prefer to just remove code instead of commenting

Tools/Adapters/BOSSAdapter.cpp
```diff
@@ -21,7 +21,7 @@ BOSS::createHelicityModel(const char *modelXMLFile, int seed,
                           const std::vector<int> &finalState,
                           const char *particleXMLFile) {
   std::string modelStr;
-  if (modelXMLFile) {
+  if (modelXMLFile == NULL) {
```
spflueger: `nullptr` would be better

Examples/SimFit/SimFit.cpp
spflueger: Could you remove the comment. I prefer to just remove code instead of commenting

PR #215, opened 2019-09-18
Fix bug in FunctionTree destructor

weidenka:
Update DalitzAnalysis_D03K example


Data/DataSet.hpp
```diff
@@ -44,6 +44,10 @@ DataSet convertEventsToDataSet(std::vector<Event>::const_iterator EventsBegin,
 
 DataSet convertEventsToDataSet(const std::vector<Event> &Events,
                                const ComPWA::Kinematics &Kinematics);
+                               
+std::vector<DataPoint>
+convertEventsToDataPoints(const std::vector<Event> &Events,
```
spflueger: What is this needed for? As far as I see nowhere. I would not put in unused/unnecessary code.

spflueger: What is this needed for? As far as I see nowhere. I would not put in unused/unnecessary code.

PR #216, opened 2019-09-19
Refactored IntensityBuilderXML

spflueger:
- A HelicityKinematics instance can now be constructed by a namespace
  function.
- The constructor of the Builder class has changed and needs a particle
  list, a Kinematics instance, and a model property tree.
- The IntensityBuilderState helper struct is now hidden to the user.


spflueger:
Could someone tell me if this change would be ok?

PR #219, opened 2019-09-24
Particlelist rework

### LINKED ISSUE #145: Single source for the default particles
Use and xml file to define the default particles. This can be read in and used by the python and c++ code. Currently the two versions being maintained is not good

spflueger:
- particle_list.xml was moved to Physics
- Closes #145
- The ParticeList.hpp file was removed. It contained a separate version
  of the particle properties as a std::string in the xml format. These
  properties are now read in from the xml file, which is the single
  source for this information.
- The shared_ptr usage for the ParticleList container was removed, since
  the use of copy, move, and references is perfectly fine.
- The ParticleList container is now a std::set and an appropriate "<"
  operator was defined. The PID currently the unique identification of
  the particle and used in the sorting.
- Refactored the ParticleList insert and read functions. Now the code is
  more simple and easier to use.
- Fixed python code to correctly interface the new read and insert
  functions and the user code calling them.
- Also introduced a search path list variable in the default settings
  python module. This helps to find the particle_list.xml file
  automatically.


weidenka:
I think the search path list in the python module is not nessesary if `particle_list.xml` is added to the python module.

spflueger:
> Deleting `ParticleList.hpp` is fine for me as long as you ensure that every example and every test runs independently from the working directory after build as well as after the install stage without the need for environment variables.

This is the case. Where do you think its not working?

spflueger:
> I think the search path list in the python module is not nessesary if `particle_list.xml` is added to the python module.

Can you clarify what you mean by this?

weidenka:
How does `DalitzAnalysis_D03K` finds `particle_list.xml`? If you run the binary from an arbitrary working directory?

spflueger:
> How does `DalitzAnalysis_D03K` finds `particle_list.xml`? If you run the binary from an arbitrary working directory?

You hand the particle list as a runtime argument

weidenka:
> > How does `DalitzAnalysis_D03K` finds `particle_list.xml`? If you run the binary from an arbitrary working directory?
> 
> You hand the particle list as a runtime argument

Than you need to know where it is. And the path is different if ComPWA in only build or installed. I think it is very important that Examples run without problems and with as little as possible sources of mistakes.

spflueger:
> > > How does `DalitzAnalysis_D03K` finds `particle_list.xml`? If you run the binary from an arbitrary working directory?
> > 
> > 
> > You hand the particle list as a runtime argument
> 
> Than you need to know where it is. And the path is different if ComPWA in only build or installed. I think it is very important that Examples run without problems and with as little as possible sources of mistakes.

Ok, let me make some things clear. It is of **utmost importance** that there is only a single default particle list (defined in a file), and all code can only parse this file and use the values in a immutable way. This requirement is far more important than making an example work anywhere in the system without having to specifiy a file etc.

Now to make your example work, you have several possibilities.
1. use hardcoded search paths or a predefined configuration location e.g. `$HOME/.ComPWA`
2. use an environment variable
3. use a runtime argument
4. maybe more?

I would prefer option 1 or 3

weidenka:
> I would prefer option 1 or 3

None of the options is a fail safe solution for the examples. We might discuss this next Monday from person to person.

spflueger:
> None of the options is a fail safe solution for the examples. We might discuss this next Monday from person to person.

To me there is nothing more to discuss. I stressed the important point/requirement, and to me this is non-negotiable. What is your standpoint on this @zhang-jingqing

weidenka:
As a compromise it would be fine for me if `DalitzFitApp` uses a hard coded particle list. `DalitzAnalysis_D03K` was not a good example since it also uses a local model file.

spflueger:
> As a compromise it would be fine for me if `DalitzFitApp` uses a hard coded particle list. `DalitzAnalysis_D03K` was not a good example since it also uses a local model file.

That is fine with me.

Just to refresh my memory, what is the reason for this app? As far as I remember its supposed to show how a more complete analysis would look like, but with a more or less arbitrary model and data. Right?
If that is the case, and each analysis is a bit different, such a generalized analysis application does not make much sense to me. Then it would simply serve as a "template" and users would use it as a starting point/reference. Consequently the general form is not needed and just using a predefined model etc would be enough.
If such kind of Dalitz analysis would be very similar for various channels etc, such a generalized application would make sense. However I would expect all of the steering to be done by command line arguments (where also the particle_list.xml would be listed). This is a perfectly normal behavior to me.

spflueger:
What is the status here? Should `DalitzFitApp` get a hardcoded particle list or get the particle list from a command line argument?

weidenka:
`DalitzFitApp` should have a hard coded particle list

spflueger:
> `DalitzFitApp` should have a hard coded particle list

Ok, so `DalitzFitApp` serves an example for a specific analysis? Then I can drop the default particle list completely right? Also, can we merge `DalitzFitApp` with the `FitTest` application to avoid duplicate code?

If not, I would recommend using a command line argument for the default particle list

weidenka:
`DalitzFit` should work in any case without the need to know the path to a default particle list. Therefore, no command line argument, no environment variable. `FitTest` is a unittest and actually not meant to be a code example. We should not spoil examples with the boost test clutter, I would therefore suggest to not merge both

spflueger:
> `DalitzFit` should work in any case without the need to know the path to a default particle list. 
Therefore, no command line argument, no environment variable.

So then I remove the default particle list code completely, add all needed particles to the hardcoded string and simply use that? 

> `FitTest` is a unittest and actually not meant to be a code example. We should not spoil examples with the boost test clutter, I would therefore suggest to not merge both

True.

PR #222, opened 2019-09-24
Increased compatibility for boost and cmake

### LINKED ISSUE #220: Increase cmake boost version compatibility
**Describe the bug**
When using boost 1.70.0 (probably also above), the dynamic linking to boost libraries does not work anymore.

Boost comes with its own CMake config code now, which changes the previous default settings. `set(BUILD_SHARED_LIBS ON)` has to be used now.

**To Reproduce**
Try to compile ComPWA with boost 1.70.0

spflueger:
- Fixes #220
- Removed ComPWA internal FindBoost.cmake module, which was incompatible
  with newer boost versions. The FindBoost module shipped with CMake is
  used instead.


spflueger:
> I think it is not a good idea the remove the FindBoost.cmake. As I explained in issue #220 to my understanding you need to upgrade cmake every time a new boost version is released. In case the method how cmake finds boost has changed: What is the minimum requirement on cmake to find the newest boost version?

Ok first of all you don't need to update cmake every time a new boost version is released. There might be problems, if they make non compatible changes.
If this happens then you need to update cmake. However we would also need to update our own FindBoost to make it work with the newest Boost again.

So I do not see the point carrying around our own FindBoost cmake file

weidenka:
> > I think it is not a good idea the remove the FindBoost.cmake. As I explained in issue #220 to my understanding you need to upgrade cmake every time a new boost version is released. In case the method how cmake finds boost has changed: What is the minimum requirement on cmake to find the newest boost version?
> 
> Ok first of all you don't need to update cmake every time a new boost version is released. There might be problems, if they make non compatible changes.
That was the case in the past. I'm not sure about the current situation.
> If this happens then you need to update cmake. However we would also need to update our own FindBoost to make it work with the newest Boost again.

> So I do not see the point carrying around our own FindBoost cmake file
We set minimum requirements on cmake and boost in the documentation. If ComPWA work with all (or at least most) combinations of cmake and boost that match those requirements I'm fine with removing the local `FindBoost.cmake`. As I said I do not think that this is the case. You might want to have a look in the official `FindBoost.cmake` of cmake. There, you have `if...else` stagements for each version of boost so I guess this needs to be updated for each new `boost` version.

spflueger:
> > > I think it is not a good idea the remove the FindBoost.cmake. As I explained in issue #220 to my understanding you need to upgrade cmake every time a new boost version is released. In case the method how cmake finds boost has changed: What is the minimum requirement on cmake to find the newest boost version?
> > 
> > 
> > Ok first of all you don't need to update cmake every time a new boost version is released. There might be problems, if they make non compatible changes.
> 
> That was the case in the past. I'm not sure about the current situation.
> 
Well with the patch I uploaded it is also working with the new boost and various verions of cmake. So far I could not get it compiling without modifying our `CMakeLists.txt file`, no matter which version of cmake.

> > If this happens then you need to update cmake. However we would also need to update our own FindBoost to make it work with the newest Boost again.
> 
> > So I do not see the point carrying around our own FindBoost cmake file
> 
> We set minimum requirements on cmake and boost in the documentation. If ComPWA work with all (or at least most) combinations of cmake and boost that match those requirements I'm fine with removing the local `FindBoost.cmake`. As I said I do not think that this is the case. You might want to have a look in the official `FindBoost.cmake` of cmake. There, you have `if...else` stagements for each version of boost so I guess this needs to be updated for each new `boost` version.

Yes they have these if else statements for various versions. However if a new boost version is the same as the old one, there will be no problems running it with on old FindBoost.

The only thing you achieve by using your own FindBoost (from a new cmake version), is that you can make old versions of cmake find boost like a new version of cmake (I'm not even sure this actually works atm, but thats beside the point). However you cannot guarantee that the versions are compatible, because maybe the new FindBoost uses cmake commands only known to a newer one etc.

The compatibility of cmake and boost has nothing to do with ComPWA. CMake will warn the user if he uses a too new version of boost.

spflueger:
The decision when and if to remove the `FindBoost.cmake` file of ComPWA is postponed to a later time. Maybe this is pull request and issue #220 are helpful for users experiencing compatibility issues with cmake and boost

PR #223, opened 2019-09-26
Geneva fix

### LINKED ISSUE #221: Fix Geneva module to compile with newest code base
The Geneva optimizer module does currently not compile

spflueger:
Fixed Geneva interface module

- Closes #221
- Geneva interface code now compiles.
- Geneva test are working (find correct minima). However the performance
  is extremely poor. Not sure what causes this.
- Adapted GenevaResult. However error estimates, covariance matrix and
  further properties are missing.
- The user can now specify an order of algorithms in the GenevaIF. These
  algorithms will be used in the minimization in exactly this order.


spflueger:
The pull request is not completely finished. The Estimator pointer has still to be fixed. But apart from this it's done.

PR #224, opened 2019-09-30
Update pybind to release 2.4.2

weidenka:
New version fixes compile time errors with AppleClang 11.


spflueger:
I just noticed that the our pybind is not a submodule, but a copy. Could you also change that?

weidenka:
Travis CI seems to got stuck. The build have passed but the status is still pending.

PR #225, opened 2019-10-01
Remove shallow cloneing of submodules

weidenka:
Shallow cloning allows only to clone relatively new commits


weidenka:
Does not work for tbb currently.

PR #227, opened 2019-10-02
Move pycompwa to a separate repository

weidenka:
* Adjust build system of ComPWA to build static libraries as well ( `-DBUILD_STATIS=ON`)
* Disable pycompwa module
* Move pycompwa code to a new repository https://github.com/ComPWA/pycompwa


.travis.yml
```diff
@@ -144,6 +163,3 @@ script:
       ctest -j 2 --output-on-failure;
     fi
   - make install
-  # Run python tests and test module installation
-  - cd pycompwa && python setup.py pytest && cd ..
-  - pip install ./pycompwa --user
```
spflueger: Has to got to pycompwa travis.yml

CMakeLists.txt
```diff
@@ -135,21 +144,6 @@ find_package( Boost COMPONENTS
   REQUIRED
```
spflueger: I would have thought that the pycompwa project would not need boost, root, etc directly, but the CMake file of ComPWA will require that. I guess these are modifications for the future

.travis.yml
```diff
@@ -82,6 +82,25 @@ matrix:
         # FindPython3.cmake finds always this version (no matter 
         # if it is activated or not)
         - pyenv global 3.7.1
+    - os: linux
+      dist: xenial
+      compiler: gcc
+      env: 
+        - TASK="gcc7 ROOT v6.16"
```
spflueger: maybe add some static label to identify the job on travis ci easier

CMakeLists.txt
weidenka: Since pycompwa builds the entire ComPWA is requires `boost`, `cmake` etc.

weidenka: I removed those requriements since they are automatically passed to target `ui` via the other ComPWA targets. Anyway in order to build pycompwa, `boost`, `cmake`, and `root` are required.

PR #243, opened 2019-10-29
Remove Minuit2 dependency from interface

weidenka:


Optimizer/Minuit2/CMakeLists.txt
```diff
@@ -1,16 +1,17 @@
 set(lib_srcs
-	MinuitFcn.cpp MinuitIF.cpp MinuitResult.cpp
+	MinuitIF.cpp MinuitResult.cpp
 )
 set(lib_headers
-	MinuitFcn.hpp MinuitIF.hpp MinuitResult.hpp
```
spflueger: Isnt this needed for XCode to get the dependencies right?

Optimizer/Minuit2/MinuitIF.hpp
```diff
@@ -30,36 +22,29 @@ namespace Minuit2 {
...
+  std::string MinuitStrategy = "medium";
+
+  // Minuit configuration variables
+  unsigned int GradientNCycles = 0;
+  double GradientStepTolerance = 0;
+  double GradientTolerance = 0;
+  unsigned int HessianNCycles = 0;
+  unsigned int HessianGradientNCycles = 0;
+  double HessianStepTolerance = 0;
+  double HessianG2Tolerance = 0;
```
spflueger: Generally it would be a bit easier UI if these members would be initialized to some standard values ("medium") as given in the `Minuit2Strategy`, and that these members are actually used in the minimization. The user can then modifiy them individually, or with some helper function which would set these values according to the string specification. Also the `MinuitStrategy` member would be dropped.

Because in this interface the user might get confused about the priority of the members, so that these variables will overwrite previous settings. 

A pity that using the `Minuit2Strategy` as a member makes the ROOT::Minuit2 library a PUBLIC dependency again...

PR #245, opened 2019-10-30
Switched to TChain instead of TTree for loading multiple files

redeboer:
Note that [`TChain::Add`](https://root.cern.ch/doc/master/classTChain.html#a78a896924ac6c7d3691b7e013bcbfb1c) takes wildcards

### DIFF
--- a/Data/Root/RootDataIO.cpp
+++ b/Data/Root/RootDataIO.cpp
@@ -5,12 +5,12 @@
 // Root-Headers
 #include "RootDataIO.hpp"
 
+#include "TChain.h"
 #include "TClonesArray.h"
 #include "TFile.h"
 #include "TLorentzVector.h"
 #include "TParticle.h"
 #include "TParticlePDG.h"
-#include "TTree.h"
 
 #include "Core/Generator.hpp"
 #include "Core/Kinematics.hpp"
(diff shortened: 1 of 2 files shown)


Data/Root/RootDataIO.cpp
```diff
@@ -26,39 +26,38 @@ RootDataIO::RootDataIO(const std::string TreeName_, int NumberEventsToProcess_)
...
+  if (!chain->GetListOfFiles()->GetEntriesFast())
     throw std::runtime_error("RootDataIO::RootDataIO() | "
-                             "Can't open data file: " +
+                             "Unable to load files: " +
                              InputFilePath);
 
-  TTree *fTree = (TTree *)File.Get(TreeName.c_str());
-
-  if (!fTree)
+  if (!chain->GetEntriesFast())
```
spflueger: You don't check the chain for nullptr here. So this line will segfault, if the tree can't be read. So you should use `if(!chain)`. Maybe it would also be better not to throw an error, but just log an error. And return an empty vector.

Data/Root/RootDataIO.cpp
```diff
@@ -78,7 +77,7 @@ RootDataIO::readData(const std::string &InputFilePath) const {
 
     Events.push_back(evt);
   } // end event loop
-  File.Close();
+  delete chain;
```
spflueger: Root will not try to delete this again? You could also try to use TChain like this `TChain chain(...);` to avoid the new in the first place.

redeboer: Hehe, it looks terrible, I know. But this is ROOT... for large files, you need to allocate it dynamically. Actually, there will be a warning message at run-time if you don't (let me look up some documentation on that issue). A `unique_ptr` is a small remedy in this case.

And yes, `delete` is fine with ROOT, it won't delete again.

Data/Root/RootDataIO.cpp
redeboer: No, `TChain` works differently than `TTree` in this regard: you don't `Get` a `TChain` from a `TFile`, but (much like `RootDataIO`) define before hand what the name of the `TTree` is when you construct the `TChain`, then `Add` files.

Data/Root/RootDataIO.cpp
spflueger: Ok. Yeah it would be fine to use pointers here. I just remember I had troubles sometimes when I deleted the pointer myself, because root's garbage collection tried to do that again at the end of the program...

Data/Root/RootDataIO.hpp
```diff
@@ -29,8 +29,9 @@ class RootDataIO {
   /// \param TreeName_	Name of tree in input or output file
   /// \param NumberEventsToProcess_	-1 processes all events
   RootDataIO(const std::string TreeName_ = "data",
-             int NumberEventsToProcess_ = -1);
+             std::size_t NumberEventsToProcess_ = -1);
```
spflueger: I just noticed this. It will work because it will set it to the max unsigned number. If we would use the max value for `size_t`, it would be a bit more transparent. `std::numeric_limits<size_t>::max()` should do it https://en.cppreference.com/w/cpp/types/numeric_limits. Then you could also remove the `NumberEventsToProcess <= 0` in the above if statement

PR #248, opened 2019-10-31
readParticles overload for ptree

redeboer:
Small changes:
* Clang-format now aligns assignment statements
* In order to make the C++ example files a bit more consistent, `readParticles` now takes a boost `ptree`, just like `createHelicityKinematics` and `IntensityBuilderXML`.

### DIFF
--- a/Core/Properties.cpp
+++ b/Core/Properties.cpp
@@ -151,16 +151,16 @@ void insertParticles(ParticleList &list, std::string FileName) {
 ParticleList readParticles(std::stringstream &Stream) {
   boost::property_tree::ptree tree;
   boost::property_tree::xml_parser::read_xml(Stream, tree);
-
-  ParticleList list;
-  insertParticles(list, tree);
-  return list;
+  return readParticles(tree);
 }
 
 ParticleList readParticles(std::string FileName) {
   boost::property_tree::ptree tree;
   boost::property_tree::xml_parser::read_xml(FileName, tree);
+  return readParticles(tree);
+}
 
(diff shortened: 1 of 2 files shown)


redeboer:
No, `AlignConsecutiveAssignments` is not, but it's one of the options I like. That's why I ask your review: it's a small modification to our coding conventions.

I haven't applied the new clang-format to other parts of the framework, because I don't want to interfere with other pull requests.

There are some problems in the rest of the code btw, will look at them later.

weidenka:
As I said to my opinion its not really unnessary and makes less simple to describe our coding convention: "We use LLVM, but with Point1 and Point2 but we exclude Point3....". But if you insist on it you can merge the request.

redeboer:
Ah, I forgot we explicitly refer to LLVM.
But the description for any new standard can be simpler: just refer to the `.clang-format` file in our repository, it speaks for itself. In practice, you should automise your code conventions anyway, not having to think about them.

In addition, if you refer to a `.clang-format`, you can update your convention at a later stage. That one rule I added is just one I thought of, but there are others we may want to implement later.

At any rate, maybe it's better to split this proposal into a separate merge request, then we can decide later.

redeboer:
Sorry, had pushed the corrections to the wrong branch, now the Travis shouldn't complain anymore.

The clang-format proposal has bene removed and will be put in another PR.

redeboer:
> Sorry, had pushed the corrections to the wrong branch, now the Travis shouldn't complain anymore.
> 
> The clang-format proposal has bene removed and will be put in another PR.

See #249

PR #249, opened 2019-11-01
Applied LLVM format standards

redeboer:
* LLVM formatting guidelines have been written explicitly to the `.clang-format` file. Currently, all rules there follow LLVM, but we can specify individual rules later.
* `CONTRIBUTE.md` refers to the `.clang-format` file.
* The LLVM standard has been applied to all source code through [ClangFormat 6](https://releases.llvm.org/6.0.0/tools/clang/docs/ClangFormatStyleOptions.html).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,1 +1,18 @@
-BasedOnStyle: LLVM
+# This file was originally created with clang-format 6.0.0 through
+# clang-format -style=llvm -dump-config > .clang-format
+# Options that agree with the LLVM standards have been removed from the resulting file
+
+# In, Linux, you can format all C++ code in this framework from the main directory of this repository using:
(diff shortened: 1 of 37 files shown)


.clang-format
```diff
@@ -1 +1,124 @@
...
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+TabWidth:        8
+UseTab:          Never
```
spflueger: I would use LLVM as a default, and overwrite these conventions/settings with only a few lines below

PR #238, opened 2019-10-24
Fit fraction rework

spflueger:
This update brings back the fit fraction functionality! Now various fit fractions can be constructed, based on the name of the Amplitudes or Intensities. Some example code on how to use the new fit fraction functionality can be seen in the DalitzFitApp.

I made some rough comparison of the fit fraction results of the DalitzFitApp. But since a lot of things have changed since the fit fraction functionality was working, it is almost impossible to get a 100% match.

Let me know what you think. I will change the commit messages once we are closer to finalizing the pull request.

PR #253, opened 2019-11-07
Dalitz analysis fit fraction

weidenka:
Extends PR #238. There is still an issue with the fit. The phi(1020) magnitude is not fitted properly. I'm not sure if that has to do with the changes to the `IntensityBuilderXML`.


spflueger:
Ok, I will have to take a look tomorrow.
>The phi(1020) magnitude is not fitted properly.

What does this mean? The value is just not within the errors?

Btw, I was not aware that any of you are using this pull request for development. So I rebased it again... So your commits probably don't match mine and you would have to fix that.

weidenka:
Problems with the fit parameters were related to a mistake in the model file.

spflueger:
> Problems with the fit parameters were related to a mistake in the model file.

Ok good

weidenka:
There is still an issue with updating the intensity from a fit result. Since `ComPWA::initializeWithFitResult()` tries to set also values of fixed parameters an expection is thrown.

Tools/Plotting/DalitzPlot.cpp
```diff
@@ -65,9 +65,9 @@ void DalitzPlot::fill(const std::vector<ComPWA::Event> &data, bool normalize,
   _plotHistograms.push_back(std::move(hist));
 }
 
-void DalitzPlot::fill(const std::vector<ComPWA::Event> &data,
-                      FunctionTreeIntensity &intens, bool normalize,
-                      std::string name, std::string title, Color_t color) {
+void DalitzPlot::fill(const std::vector<ComPWA::Event> &data, Intensity &intens,
+                      bool normalize, std::string name, std::string title,
```
redeboer: Why no `const std::string&`?

weidenka: For simply opjects like `std::string` or `double` I doubt that there is any difference after compiler optimizations. For larger objects like `const std::vector<ComPWA::Event> &data` one should certainly pass by reference.

redeboer: Yeah, but just a convention thing
https://github.com/ComPWA/ComPWA/blob/master/CONTRIBUTING.md#const-correctness

redeboer: For `double` a reference would actually be slower btw because of dereferencing, and the size of the pointer is also the same

PR #271, opened 2020-01-16
Added ROOT_DIR option for cmake

redeboer:
Currently, ComPWA requires you to have sourced `thisroot.sh` of your ROOT installation, or set `ROOTSYS` in order to `find_package(ROOT)`. If you have several installations of ROOT, this could be problematic, and in general it's not so clean to require users to have sourced things first.

Now you can use `cmake -DROOT_DIR=...` to specify which installation you want.

### DIFF
--- a/cmake/Modules/FindROOT.cmake
+++ b/cmake/Modules/FindROOT.cmake
@@ -21,6 +21,10 @@ if("${ROOT_DIR}" STREQUAL "" AND NOT "$ENV{ROOTSYS}" STREQUAL "")
   set(ROOT_DIR "$ENV{ROOTSYS}")
 endif()
 
+if(NOT "${ROOT_DIR}" STREQUAL "")
+  list(APPEND CMAKE_PREFIX_PATH ${ROOT_DIR})
+endif()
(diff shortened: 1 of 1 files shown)


redeboer:
[This change](https://github.com/redeboer/ComPWA/blob/6b33a9960b84ac5888f370c054a6b3789c3446ae/CMakeLists.txt#L142) is based on https://root.cern.ch/how/integrate-root-my-project-cmake, but does not seem to work properly. See also #272.

redeboer:
> [This change](https://github.com/redeboer/ComPWA/blob/6b33a9960b84ac5888f370c054a6b3789c3446ae/CMakeLists.txt#L142) is based on https://root.cern.ch/how/integrate-root-my-project-cmake, but does not seem to work properly. See also #272.

Okay, fixed by removing [`cmake/Modules/FindROOT.cmake`](https://github.com/ComPWA/ComPWA/blob/2fdf155ffacc214a8bbc9bff8984849d4b4f5eaa/cmake/Modules/FindROOT.cmake)

PR #273, opened 2020-01-17
Simplified read/write data for ROOT

redeboer:


redeboer:
This is a modified version of #268 that doesn't work with a `EventDataStructure`. Since there is not really a need for this `struct` within the two functions `readData` and `writeData` (which is what this PR is all about), this `struct` was removed.

Data/Root/RootDataIO.hpp
```diff
@@ -14,30 +14,39 @@ class TTree;
...
+/// These arrays should contain
+/// [`TParticle`](https://root.cern.ch/doc/master/classTParticle.html) objects
+/// with a defined 4-momentum.
+/// * A branch of `double`s called `"Weight"`.
+/// \param InputFileName Input ROOT file(s); can take wildcards, see
+/// [`TChain::Add`](https://root.cern.ch/doc/master/classTChain.html).
+/// \param TreeName Name of the event-based tree
+/// \param NumberEventsToRead Limit the resulting vector to this number of
+/// events (optional).
+std::vector<ComPWA::Event> readData(const std::string &InputFileName,
```
spflueger: For such functions I make the strings just via simple copy, so `std::string Input...`

Data/Root/RootDataIO.cpp
```diff
@@ -21,116 +23,115 @@ namespace ComPWA {
...
-  chain.Add(InputFilePath.c_str());
-  if (!chain.GetListOfFiles()->GetEntriesFast())
-    throw std::runtime_error("RootDataIO::RootDataIO() | "
-                             "Unable to load files: " +
-                             InputFilePath);
+std::vector<ComPWA::Event> readData(const std::string &InputFilePath,
+                                    const std::string &TreeName,
+                                    long long NumberOfEventsToRead) {
+  // Ignore custom streamer warning and error message for missing trees
+  gErrorIgnoreLevel = kFatal;
```
spflueger: This also ignores fatal errors?

Data/Root/test/RootDataIOTest.cpp
```diff
@@ -17,25 +24,183 @@
...
+    TheChain.Fill();
+  TheChain.Write();
+  TheFile.Close();
 
-  auto sampleIn(RootIO.readData("RootReaderTest-output.root"));
-  BOOST_CHECK_EQUAL(sample.size(), sampleIn.size());
+  try {
+    ComPWA::Data::Root::readData(FileName, TreeName);
+  } catch (const std::runtime_error &e) {
+    BOOST_CHECK_EQUAL(e.what(), "Root::readData() | TTree does not have a "
```
spflueger: In that case it would be better to throw more precise exceptions and compare the exception type instead of the error message

Data/Root/RootDataIO.hpp
redeboer: I see both. Anyway, just trying to follow:
https://github.com/ComPWA/ComPWA/blob/master/CONTRIBUTING.md#const-correctness

Data/Root/RootDataIO.cpp
redeboer: I increased the level a bit now, to `kBreak`, but the above means that  `kFatal` is still shown.

Data/Root/test/RootDataIOTest.cpp
redeboer: Added `CorruptFile` exception

PR #275, opened 2020-01-20
Function tree refactor

spflueger:
The FunctionTree class was removed completely since it had no
functionality whatsoever and just pointlessly wrapped around the
TreeNode.
The construction of a Tree is now more straight forward. Before unique
names of the nodes were used to built up the tree. Which was very
confusing.
Now there are no more names and you simply add TreeNode objects to other
TreeNodes.
Also a lot of unneeded methods were removed.


Physics/HelicityFormalism/WignerD.hpp
```diff
@@ -5,10 +5,10 @@
 #ifndef COMPWA_PHYSICS_HELICITY_FORMALISM_WIGNERD_HPP_
 #define COMPWA_PHYSICS_HELICITY_FORMALISM_WIGNERD_HPP_
 
-#include "Core/FunctionTree/FunctionTree.hpp"
 #include "Core/FunctionTree/Functions.hpp"
+#include "Core/FunctionTree/TreeNode.hpp"
 
-#include "qft++/WignerD.h"
+#include "ThirdParty/qft++/include/qft++/WignerD.h"
```
redeboer: Why include like this?

Tools/test/IntegrationTest.cpp
```diff
@@ -117,10 +117,10 @@ const std::string JpsiDecayKinematics = R"####(
 )####";
 
 const std::string JpsiDecayTree = R"####(
-<Intensity Class='CoherentIntensity' Name='coherent_0'>
```
redeboer: Is backward compatible, right?

Physics/Dynamics/Voigtian.hpp
```diff
@@ -97,10 +97,10 @@ inline std::complex<double> dynamicalFunction(double mSq, double mR, double wR,
   return result;
 }
 
-std::shared_ptr<ComPWA::FunctionTree::FunctionTree>
+std::shared_ptr<ComPWA::FunctionTree::TreeNode>
 createFunctionTree(InputInfo Params,
                    const ComPWA::FunctionTree::ParameterList &DataSample,
-                   unsigned int pos, std::string suffix);
```
redeboer: What was this for originally? Those unique names?

Examples/DalitzFit/DalitzFitApp.cpp
```diff
@@ -75,7 +75,7 @@ std::string amplitudeModel = R"####(
 		    </Amplitude>
       </Amplitude>
     </Amplitude>
-	  <Amplitude Class="CoefficientAmplitude" Name="myAmp">
+	  <Amplitude Class="CoefficientAmplitude" Component="myAmp">
```
redeboer: Component is a new tag?

Physics/Dynamics/NonResonant.hpp
```diff
@@ -1,24 +0,0 @@
-// Copyright (c) 2015, 2017 The ComPWA Team.
```
redeboer: Why removed?

Physics/HelicityFormalism/WignerD.hpp
spflueger: We have a slightly customized version of qft++ and its a quite strict part of ComPWA

PR #276, opened 2020-01-21
Upgraded AsciiReader

redeboer:
AsciiReader now takes a more advanced file format, see [here](https://github.com/redeboer/ComPWA/blob/AsciiReader/Data/AsciiReader/AsciiReader.hpp#L17).


Data/AsciiReader/AsciiReader.cpp
```diff
@@ -8,46 +8,165 @@
...
-AsciiReader::AsciiReader(unsigned int NumberOfParticles_)
-    : NumberOfParticles(NumberOfParticles_) {}
+std::string tolower(std::string s) {
+  std::transform(s.begin(), s.end(), s.begin(),
+                 [](unsigned char c) { return std::tolower(c); });
+  return s;
+}
 
-AsciiReader::~AsciiReader() {}
+bool IsHeaderLine(std::string line) {
```
spflueger: Use the lower case camel style for function names `isHeaderLine`

Data/AsciiReader/AsciiReader.cpp
```diff
@@ -8,46 +8,165 @@
...
+  return (tolower(line).find("header") != std::string::npos);
+}
 
-std::shared_ptr<std::vector<ComPWA::Event>>
-AsciiReader::readData(const std::string &InputFilePath) const {
+bool IsEmptyString(const std::string &line) {
```
spflueger: Same as above

PR #277, opened 2020-01-21
AsciiReader bugfix

redeboer:


Data/Ascii/test/AsciiReaderTest.cpp
```diff
@@ -62,14 +85,14 @@ BOOST_AUTO_TEST_CASE(TestMomentumEnergyOrder) {
 }
 
 BOOST_AUTO_TEST_CASE(TestOverwrite) {
-  const char *Filename = "Data_AsciiReaderTest-TestOverwrite.dat";
+  const char *FileName = "Data_AsciiReaderTest-TestOverwrite.dat";
```
spflueger: Why use a const char* here (also the std::remove below)?

Data/Ascii/AsciiReader.cpp
```diff
@@ -6,14 +6,16 @@
 #include <sstream>
 #include <utility>
 
-#include "AsciiReader.hpp"
+#include "Data/Ascii/AsciiReader.hpp"
```
spflueger: You could rename the files to AsciiIO, because you also wrote a write function now.

Data/Ascii/test/AsciiReaderTest.cpp
redeboer: Yeah `string` is better. The remove is just to clean the output after running `ctest`. Should I delete it?

Data/Ascii/AsciiReader.cpp
redeboer: Would agree better with `RootDataIO`. But the three I's are also ugly.

Data/Ascii/test/AsciiReaderTest.cpp
spflueger: Yeah it gets deleted after running out of scope anyways

redeboer: No, `remove` deletes the output file

Data/Ascii/AsciiReader.cpp
spflueger: Ok, up to you

Data/Ascii/test/AsciiReaderTest.cpp
redeboer: [`std::remove`](https://en.cppreference.com/w/cpp/io/c/remove) requires `const char*` for some reason...

Data/Ascii/AsciiReader.cpp
redeboer: Renamed to `AsciiDataIO` in accordance with `RootDataIO`

Data/Ascii/test/AsciiReaderTest.cpp
spflueger: I see, nvm then

Issue #281, opened 2020-01-29
TreeNodes will not destruct themselves


spflueger:
A FunctionTree can currently not be deleted because of the cyclic linking of the TreeNodes amongst each other (Parent <-> Daughter). Also the Observer Pattern is introducing this cyclic dependency for TreeNodes with Parameters.

**Expected effort: 2d**

PR #284, opened 2020-01-29
FunctionTree memory leak fix

spflueger:
Solves issue #281

### DIFF
--- a/Core/FunctionTree/Functions.hpp
+++ b/Core/FunctionTree/Functions.hpp
@@ -29,9 +29,9 @@ namespace FunctionTree {
 ///
 class Strategy {
 public:
-  Strategy(ParType in, std::string op = "") : checkType(in), Op(op){};
+  Strategy(ParType in, std::string name) : checkType(in), Name(name){};
 
-  virtual ~Strategy() {}
+  virtual ~Strategy() = default;
 
   /// Return parameter type
   virtual const ParType OutType() const { return checkType; }
@@ -40,7 +40,7 @@ public:
   virtual void execute(ParameterList &paras,
                        std::shared_ptr<Parameter> &out) = 0;
 
-  std::string str() const { return Op; }
+  std::string str() const { return Name; }
 
   friend std::ostream &operator<<(std::ostream &out,
                                   std::shared_ptr<Strategy> b) {
@@ -53,7 +53,7 @@ public:
 
 protected:
   ParType checkType;
-  const std::string Op;
+  const std::string Name;
 };
 
 ///
--- a/Core/FunctionTree/Parameter.hpp
+++ b/Core/FunctionTree/Parameter.hpp
@@ -82,7 +82,7 @@ public:
   Parameter(std::string name, ParType type = ParType::UNDEFINED)
       : Name(name), Type(type) {}
 
-  virtual ~Parameter() {}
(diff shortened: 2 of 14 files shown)


Core/FunctionTree/Functions.hpp
```diff
@@ -29,9 +29,9 @@ namespace FunctionTree {
 ///
 class Strategy {
 public:
-  Strategy(ParType in, std::string op = "") : checkType(in), Op(op){};
```
redeboer: Just a naming thing: not entirely clear to me what `op` is?

Core/FunctionTree/Parameter.hpp
```diff
@@ -101,30 +101,14 @@ class Parameter {
   // Observer Pattern Functions
 
   /// Attaches a new TreeNode as Observer
-  void Attach(std::shared_ptr<ParObserver> newObserver) {
+  void Attach(std::weak_ptr<ParObserver> newObserver) {
```
redeboer: So if I understand correctly, the key point is that deletion is not anymore done by all `shared_ptr` running out of scope?

Core/FunctionTree/Functions.hpp
spflueger: Hmm true, I could have also improved that. If there is some more stuff change, then I can do that. Otherwise I feel a bit lazy to update the PR just for this

redeboer: Nah is fine, just took me a bit more effort to understand the code

Core/FunctionTree/Parameter.hpp
spflueger: Hmm I can't follow your question. I try to explain the core idea shortly.
Before there where `shared_ptr` of objects that had `shared_ptr` members to each other, creating a cyclic reference. That prevents running out of scope to clean up automatically.
The solution is to use a `weak_ptr` for one of the links, since `weak_ptr` avoid increasing the `shared_ptr` count of the object. Then the automatic cleanup works.
Because all of the `weak_ptr` are only used if the objects they are pointing to actually exist (by construction), its safe to use this.
I should verify that the parameters are actually also all deleted, to ensure that the Observers are also guaranteed to exist.

redeboer: Okay thanks, yes that answers my question.

spflueger: Ok so I checked and also this part is working fine. Because when we create an Intensity, all data Parameters and fit Parameters are created completely fresh and are only stored within the Intensity. So when the Intensity object runs out of scope, it is guaranteed by design to clean up nicely without any issues and no old Parameter are ever reused.

Issues could for example occur if someone would change the code that the `PartProps` would return a `shared_ptr` of the Mass of the Particle and if that would be inserted in the FunctionTree on every new construction.

Core/FunctionTree/Functions.hpp
spflueger: Ok is fixed

Core/FunctionTree/Parameter.hpp
spflueger: The new code also adds a customized destructor for the TreeNode, which unregisters itself from all objects with weak ptrs to it.

PR #290, opened 2020-02-07
DataSet, Intensity and Helicity kinematics rework 

spflueger:
This is not completely finished yet. So some cleanup work has still to be done, but the big changes are finished.

The main parts that changed:
* DataSet now has an `unordered_map` of the individual data vectors (instead of a vector)
* The Intensity interface now also expect this map as input
* The kinematics interface was also adapted to process this new data format
* The helicity kinematics class was reworked, as foundation work for the fix of the break up momenta.
Variables are now separated into angles and invariant masses and uniquely defined by their name. The name also allows easy access from the returned DataSet.

### DIFF
--- a/Core/Event.cpp
+++ b/Core/Event.cpp
@@ -36,17 +36,4 @@ double getMaximumSampleWeight(const std::vector<Event> &sample) {
   return MaxWeight;
 }
 
-std::ostream &operator<<(std::ostream &os, const DataPoint &p) {
-  os << "(";
-  for (unsigned int i = 0; i < p.KinematicVariableList.size(); ++i) {
-    os << p.KinematicVariableList[i];
-    if (i != p.KinematicVariableList.size() - 1) {
-      os << ", ";
-    }
-  }
-  os << ")";
-
-  return os;
-}
-
 } // namespace ComPWA
--- a/Physics/EvtGen/DalitzKinematics.cpp
+++ b/Physics/EvtGen/DalitzKinematics.cpp
@@ -7,6 +7,7 @@
 #include "Core/Event.hpp"
 #include "Core/Logging.hpp"
 #include "Core/Particle.hpp"
+#include "Data/DataSet.hpp"
 
 #include "ThirdParty/qft++/include/qft++/Vector4.h"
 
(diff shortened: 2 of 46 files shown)


Core/Event.cpp
```diff
@@ -36,17 +36,17 @@ double getMaximumSampleWeight(const std::vector<Event> &sample) {
...
-    os << p.KinematicVariableList[i];
-    if (i != p.KinematicVariableList.size() - 1) {
-      os << ", ";
-    }
-  }
-  os << ")";
-
-  return os;
-}
+// std::ostream &operator<<(std::ostream &os, const DataPoint &p) {
```
redeboer: Please remove (not comment) if truly redundant

spflueger:
Ok, the PR is finished now. I apologize for the rather large PR, but the DataSet change basically triggered all of this.

Physics/EvtGen/DalitzKinematics.cpp
```diff
@@ -50,200 +51,75 @@ DalitzKinematics::DalitzKinematics(
...
+                   event.ParticleList[1].fourMomentum())
+                      .invariantMass());
+    qBC.push_back((event.ParticleList[1].fourMomentum() +
+                   event.ParticleList[2].fourMomentum())
+                      .invariantMass());
+    qCA.push_back((event.ParticleList[2].fourMomentum() +
+                   event.ParticleList[0].fourMomentum())
+                      .invariantMass());
+    weights.push_back(event.Weight);
+  }
```
weidenka: Can this still handel events which are outside phase space due to numerical instabilities? Or is this not a problem anymore?

weidenka:
Furthermore, since there were quite some changes it would be nice if you could post a log file of `DalitzAnalysis_D03K` before and after your changes.

spflueger:
>Looks good to me. You moved some logic which have been in independend functions back to Kinematics (e.g. Data::reduceToPhsp(data, kin); vs. kin.reduceToPhsp(data)). Since the move was the other way around a while ago, we should agree on some standart here.

When was that the case? I think you are confusing `isWithinPhaseSpace()` with `reduceToPhsp()`. Anyways the important difference is that now the Kinematics class processes a list of Events instead of just single Events. Hence the Kinematics interface changed: `isWithinPhaseSpace()` was removed and replaced with `reduceToPhsp()`.

Physics/EvtGen/DalitzKinematics.cpp
spflueger: How do you mean? The logic did not change here. Also before the conversion is always performed. If an Event is outside the phsp boundary, the kinematic variables returned here will show that and `isWithinPhsp()` checked exactly that.

Originally exceptions were thrown inside convert, but this was changed a while ago because parallelstl does not handle throwing code (for good reason).

This was with Event based processing in the Kinematics though. Now such a feature could theoretically be inserted again, because convert does not need to run inside a parallelstl loop anymore (pstl can be used inside convert though, which would be perfectly fine).
But as for the current code: A generator should only create events inside phsp. If there are some troubles you can run the Events through the `reduceToPhsp()` method. And convert will simply convert

spflueger:
> Furthermore, since there were quite some changes it would be nice if you could post a log file of `DalitzAnalysis_D03K` before and after your changes.

Thats a good point. I will do that

spflueger:
> > Furthermore, since there were quite some changes it would be nice if you could post a log file of `DalitzAnalysis_D03K` before and after your changes.
> 
> Thats a good point. I will do that

```diff
71,76c71,85
< Info in <TH1D::Chi2TestX>: There is a bin in h1 with less than 1 event.
< 
< Info in <TH1D::Chi2TestX>: There is a bin in h1 with less than 1 event.
< 
< Info in <TH1D::Chi2TestX>: There is a bin in h1 with less than 1 event.
< 
---
> Error in <TH1D::Chi2TestX>: one histogram is empty
> Error in <TPad::Range>: illegal world coordinates range: x1=0.869640, y1=0.000000, x2=1.921921, y2=0.000000
> Error in <TPad::RangeAxis>: illegal axis coordinates range: xmin=0.974868, ymin=0.000000, xmax=1.869307, ymax=0.000000
> Error in <TH1D::Chi2TestX>: one histogram is empty
> Error in <TPad::Range>: illegal world coordinates range: x1=0.877078, y1=0.000000, x2=1.932878, y2=0.000000
> Error in <TPad::RangeAxis>: illegal axis coordinates range: xmin=0.982658, ymin=0.000000, xmax=1.880088, ymax=0.000000
> Error in <TH1D::Chi2TestX>: one histogram is empty
> Error in <TPad::Range>: illegal world coordinates range: x1=0.877078, y1=0.000000, x2=1.932878, y2=0.000000
> Error in <TPad::RangeAxis>: illegal axis coordinates range: xmin=0.982658, ymin=0.000000, xmax=1.880088, ymax=0.000000
> Error in <TPad::Range>: illegal world coordinates range: x1=0.869640, y1=0.000000, x2=1.921921, y2=0.000000
> Error in <TPad::RangeAxis>: illegal axis coordinates range: xmin=0.974868, ymin=0.000000, xmax=1.869307, ymax=0.000000
> Error in <TPad::Range>: illegal world coordinates range: x1=0.877078, y1=0.000000, x2=1.932878, y2=0.000000
> Error in <TPad::RangeAxis>: illegal axis coordinates range: xmin=0.982658, ymin=0.000000, xmax=1.880088, ymax=0.000000
> Error in <TPad::Range>: illegal world coordinates range: x1=0.877078, y1=0.000000, x2=1.932878, y2=0.000000
> Error in <TPad::RangeAxis>: illegal axis coordinates range: xmin=0.982658, ymin=0.000000, xmax=1.880088, ymax=0.000000
80,82c89,91
```

spflueger:
Ok, the fix for the DalitzPlot is also included now.

Physics/HelicityFormalism/HelicityKinematics.cpp
```diff
@@ -119,7 +119,7 @@ HelicityKinematics::calculateHelicityAngles(const Event &Event,
 
     Recoil.Boost(DecayingState);
 
-    // rotate recoil so that recoil points in the z direction
+    // rotate recoil so that recoil points in the -z direction
```
redeboer: --> z-direction ?

spflueger: Afais `M_PI - Recoil.Theta()` puts the Recoil on the -z axis, while `-Recoil.Theta()` would put it on the z axis.

PR #291, opened 2020-02-11
docs: add git commit conventions

redeboer:


CONTRIBUTING.md
```diff
@@ -65,4 +64,11 @@ side. E.g. ``float *pf; if (NULL == pf);``
 
 ## Python Code
```
spflueger: That python related stuff should be removed I would say

PR #303, opened 2020-02-26
style: implement and apply cmake-format

redeboer:
See [cmake-format](https://github.com/cheshirekow/cmake_format), using v0.6.9.


wgradl:
This seems to make a pig's ear of some of the comments, where the line breaks actually have been placed on purpose.   Can this formatting tool be taught to leave comments alone?

redeboer:
Good point, `cmake-format` [reflows comments by default](https://cmake-format.readthedocs.io/en/latest/format-features.html#markup), I guess because it assumes they shouldn't affect CMake.

See option [`enable_markup`](https://cmake-format.readthedocs.io/en/latest/configopts.html#enable-markup), I set it to False now and reapplied, let's see if that improves things...

wgradl:
Looks good to me, at least as far as the comments are concerned.

Issue #280, opened 2020-01-27
Improve header of the ASCII reader and make compatible with Pawian


redeboer:
The current [`Ascii::readData`](https://github.com/ComPWA/ComPWA/blob/cb584cd0a661ba0c87d113e876edd168c964b30f/Data/Ascii/AsciiDataIO.hpp#L31) function can read Pawian-like momentum tuple files with and without a 'header' (see Doxygen comments for illustration of the format). However, there are a few issues with this:

* Current deduction of p,E vs E,p order of the momentum tuples is deduced by checking which of the 'order interepretation' leads to a positive invariant mass for the first momentum tuple. But this check is not a guarantee that the order interpretation is the correct one. So the order should be in the header as well.
* The header lacks units.
* `Pids: ...` would be better than `Pid: ...`,  `Pid: ...`, etc
* The format follows its own thought-up syntax, which should be avoided. Maybe formal YAML would be better (e.g. `Pids: [211, 421, -411]`). But then we would need to use a real third-party parser, so that users can actually use real YAML syntax.

To make matters more complicated, there is interest to have this header parser implemented in Pawian as well. Since the *aim of this reader is to make comparison between the two frameworks more easy*, decisions on this issue should be fine-tuned to each other.

redeboer:
Note: there was an idea to allow the user to tell the `Ascii::readData` function how to interpret the data file by feeding it a `Kinematics` object (which contains the final state particles). Considering the remarks above, it's probably better to abandon that idea and to expect the user to supply a header in the data file.

spflueger:
Yeah the header solution is better. Are there any news on this?

redeboer:
![](https://github.trello.services/images/mini-trello-icon.png) [ASCII reader](https://trello.com/c/Pc8OMGkJ/9-ascii-reader)

redeboer:
Fixed by https://github.com/ComPWA/ComPWA/pull/304

PR #304, opened 2020-03-03
feat!: add and implement AsciiHeader object

redeboer:
This PR changes the syntax of ASCII input file: they now require a header section (surrounded by a `header` opening and closing tag) in YAML syntax.


Data/Ascii/AsciiDataIO.cpp
```diff
@@ -114,53 +94,62 @@ EventCollection readData(const std::string &InputFilePath,
...
-      if (IsOrderEnergyMomentum)
+      if (IsEnergyFirst)
         InputStream >> e >> px >> py >> pz;
       else
         InputStream >> px >> py >> pz >> e;
       FourVectors.push_back(ComPWA::FourMomentum(px, py, pz, e));
     }
     if (!InputStream.fail())
-      EvtList.Events.push_back(Event{FourVectors, weight});
+      ImportedData.Events.push_back(Event{FourVectors, weight});
```
spflueger: Where do you read in the weight?

Data/Ascii/AsciiHeaderIO.cpp
```diff
@@ -0,0 +1,140 @@
...
+  std::stringstream ss(line);
+  float value;
+  return (bool)(ss >> value);
+}
+
+/// @endcond
+
+std::string AsciiHeader::extractHeaderContent(std::istream &InputStream) {
+  /// -# Define regular expressions
+  static const std::regex HeaderBegin("<header>", std::regex_constants::icase);
```
spflueger: I thought its yaml style? So, `Header:`. Why you support <header>?

redeboer: It was decided in the last meeting that we stack the yaml between open and closing tags (html/xml) in this case. The file in itself is not yaml anyway because of the data outside the header section.

Data/Ascii/AsciiDataIO.cpp
spflueger: oh yeah.. ups

Data/Ascii/AsciiHeaderIO.cpp
spflueger: But some tests use `Header:`, so that is also supported?

spflueger: Ah no I missed that. They all have been changed. Ok fine with me

redeboer: No, `Header` is not supported as begin/end tag, so the test files were modified as well, but I noticed a mistake in the doxygen string of AsciiDataIO.hpp now...

redeboer:
Closes #280

PR #306, opened 2020-03-06
Eventlist refactor

redeboer:
In the `Event` class, `Particle` has been replaced by `FourMomentum`. The particle information is contained within a list of PIDs that is contained within the `EventList` class. This required a lot of refactoring, but allows for checking the PIDs when for instance doing kinematic variable conversion.

Note that the format of RootDataIO has been changed as well: no use is made of the infamous `TClonesArray` anymore (note that [even ROOT writes](https://root.cern.ch/doc/master/classTTree.html) "It is recommended to use STL containers instead of `TClonesArrays`").

I also took the liberty to rename several variables to make them agree with our naming conventions.


Core/Kinematics.hpp
```diff
@@ -16,8 +16,7 @@ namespace Data {
 struct DataSet;
 }
 
-/// The Kinematics interface is responsible for converting an Event into a
-/// DataPoint.
+/// The Kinematics interface defines the conversion of Events to a DataSet.
 class Kinematics {
```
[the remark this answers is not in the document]
redeboer: Okay will try

redeboer:
The [design of `EventCollection`](https://github.com/ComPWA/ComPWA/pull/306/files#diff-2ff3123a16ee1e8714edd124396ca852R25) could definitely be made safer (either making `Events` and `Pids` `const` or by making them private). I tried but then ran into several problems where (1) STL algorithms need direct access to the `Events` vector and (2) the `EventCollection` is build up in steps using for instance `push_back`.

Data/Ascii/AsciiDataIO.hpp
```diff
@@ -28,11 +28,11 @@ namespace Ascii {
 /// \f$\pi^-\f$, then one for the \f$\gamma\f$, and finally back to \f$\pi^+\f$.
 /// You may choose to start each event group with a weight value, but you don't
 /// need to.
-std::vector<ComPWA::Event> readData(const std::string &InputFilePath,
-                                    long long NumberEventsToRead = -1);
+EventCollection readData(const std::string &InputFilePath,
```
spflueger: Just a small remark (just in case you are not aware of it): I (try to) use arguments by copy in case the code is non performance critical, and in many cases the compiler can optimize the copy with a move anyways.

Data/Generate.cpp
```diff
@@ -263,20 +274,19 @@ generate(unsigned int NumberOfEvents, const ComPWA::Kinematics &Kinematics,
...
-  return events;
+  return GeneratedEvents;
 }
 
-std::vector<ComPWA::Event>
-generatePhsp(unsigned int nEvents,
-             const ComPWA::PhaseSpaceEventGenerator &Generator,
-             ComPWA::UniformRealNumberGenerator &RandomGenerator) {
-  std::vector<ComPWA::Event> sample;
+EventCollection generatePhsp(unsigned int nEvents, std::vector<pid> Pids,
```
spflueger: Why can't the generator interface be changed to return EventCollections? I believe this would make things easier

redeboer: It does now, right?

spflueger: I mean this interface "Core/Generator.hpp"

redeboer: There is no mention of `std::vector<Event>` in `Core/Generator.hpp`

spflueger: Thats the problem. I believe it should be returning an EventCollection

redeboer: All functions in `Generator.hpp` are returning `EventCollection`s

PR #310, opened 2020-03-18
fix: remove faulty double p_y print

redeboer:


Core/test/ParticleTest.cpp
```diff
@@ -4,8 +4,8 @@
```
spflueger: You could also rename this test to FourMomentumTest (also the TestSuite thing below)

redeboer: Good one, see https://github.com/ComPWA/ComPWA/pull/310/commits/66b5656117a3a1286c270b34e7da7cb1d2a2d399

PR #312, opened 2020-05-06
docs: add exhale documentation for Read the Docs

redeboer:
API is now built through Sphinx in addition to the original Doxygen build. This is to facilitate `intersphinx` within the four repositories that we now have (overview at [pwa.readthedocs.io](https://pwa.readthedocs.io/README.html)).

- Output can be seen [here](https://pwa.readthedocs.io/projects/compwa/en/latest/).
- The old documentation at [compwa.github.io/ComPWA](https://compwa.github.io/ComPWA/) will still be available.


.travis.yml
```diff
@@ -24,29 +25,35 @@ addons:
   apt:
     update: true
     sources:
-      - ubuntu-toolchain-r-test # gcc up to 8.0
+      - ubuntu-toolchain-r-test
       - llvm-toolchain-xenial-8
```
spflueger, line 29: shouldn't this be added to only the xenial machines (default is bionic now)

doc/Doxyfile-doxygen
```diff
@@ -0,0 +1,2432 @@
+# Doxyfile 1.8.13
```
spflueger, line 1: Its really hard to review this. My initial question was: Why is this needed? I guess its to output the info to xml? Turns out its the other way around now. Anyways I would do it this way:
1. leave the normal doxygen file for doxygen
2. then change this filename to e.g. Doxyfile-rtd and use `@INCLUDE = config_file_name` and just overwrite values (see http://www.doxygen.nl/manual/config.html). So just use the normal doxyfile as a base and just modify a few lines that you want to change.

This makes it really easy to understand and keeps things synchronized (we would like both docs to show the highest degree of identical content, right).

doc/Doxyfile
```diff
@@ -780,9 +790,7 @@ WARN_LOGFILE           =
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = ./
-INPUT                 += doc/main.md
-USE_MDFILE_AS_MAINPAGE = doc/main.md
```
spflueger, line 785: doesn't this remove the main.md from the doxygen?

.travis.yml
```diff
@@ -1,10 +1,11 @@
 language: cpp
 os: linux
+dist: bionic
```
spflueger, line 3: All machines are set to xenial again though. We could use 1 or 2 bionic jobs

doc/main.md
```diff
@@ -1,8 +0,0 @@
-Common Partial-Wave-Analysis Framework (ComPWA) {#mainpage}
-===========================================================
-
-Welcome to the class documentation of ComPWA!
-
-These pages are useful if you are working on the C++ backend of the %ComPWA framework. See [the repository on GitHub](https://github.com/ComPWA/ComPWA) for the source code.
-
-ComPWA is best used through it's python interface (`pycompwa`). You can find all installation instructions and usage examples on our website, [compwa.github.io](https://compwa.github.io/).
```
spflueger, line 8: I currently dont understand what the doxygen documentation will show as a mainpage.

doc/requirements.txt
```diff
@@ -0,0 +1,3 @@
+exhale
```
spflueger, line 1: breathe not needed?

doc/main.md
redeboer: Ah good point, sorry that was removed in an earlier version

doc/requirements.txt
redeboer: Nope, breathe is a dependency of exhale

doc/main.md
redeboer: https://github.com/ComPWA/ComPWA/pull/312/commits/89d866caf8a517870eda5f2a5150843d863248ad

.travis.yml
redeboer: Yeah xenial is a bit outfashioned, which is also why we upgrade `gcc` in those jobs. This travis file has to be completely fixed anyway...

doc/Doxyfile
redeboer: Yeah but this page is not used by Sphinx-exhale, it uses `index.rst`.

.travis.yml
redeboer: Hm, I just didn't like the comments, should they be added back? See https://github.com/ComPWA/ComPWA/pull/312#discussion_r420625184 though.

doc/Doxyfile-doxygen
redeboer: I don't like it either. Unfortunately though, Exhale requires a file named `Doxyfile` and its settings need to be slightly different than the original `Doxyfile` (for instance, because `exhale` uses XML, not HTML output). This is why there are two versions of the Doxyfile now.

I didn't know about the Include thing, I'll look into it.

.travis.yml
spflueger: I'm referring to the `llvm-toolchain-xenial-8`. I don't know what the default os for travis is, but our default is set to bionic now. I'm not sure if bionic machines would just skip this package or throw an error. Since we do not have any bionic machines we don't really know what happens (hence maybe also add 1 or 2 bionics)

redeboer: Not sure, perhaps better to move this to a separate issue

spflueger: All right.

doc/Doxyfile-doxygen
spflueger: https://exhale.readthedocs.io/en/latest/reference/configs.html#exhale.configs.exhaleDoxygenStdin

